PR #15, opened 2014-12-08
Make things radically simpler

jdeal:
@bcooksey @bryanhelmig Toying with some ideas to rip out anything strange and make formatic a more straightforward react component.

Basic usage would just be:

```js
var Form = React.createFactory(Formatic);

React.render(Form({
  config: {...},
  fields: [
    {label: 'Name', type: 'string', key: 'name'}
  ],
  value: {name: 'Joe Foo'},
  onChange: function (newValue, info) {
    console.log(newValue, info);
  }
}), document.getElementById('user'));
```

Which is pretty much by the book react.

Anything configurable would just be passed in via the `config` object. That would be passed around to all child components in props. This could include additional components for types, helper functions like humanizing, whatever. Components would then just explicitly use the config. This would add a bit more verbosity, but in exchange, things would be a lot simpler to follow.

Any compilation of fields (if even necessary) would probably just be a static method like:

```js
var fields = Formatic.compileFields([
    {label: 'Name', type: 'text', key: 'name'}
]);

// Now pass compiled fields to factory as above.
```

(Above is very simple compilation where types mapped.)

Right now, thinking about plugins. In the sense like bootstrap that can affect many components. Thinking perhaps plugins could just be functions that modify a configuration. So, for example, the bootstrap plugin could just add class names to the config to be picked up by components. Of course plugins would have to be well-behaved since they'd have access to the entire config. But again, they'd be conceptually dead simple. Usage could be a static method something like:

```js
var myPlugin = function (config) {
  config.someThings = config.someThings || [];
  config.someThings.push({...});
};

var config = Formatic.config(Formatic.plugins.bootstrap, myPlugin, {someFlag: true});

// Now pass config to factory as above.
```

Also will need to think about dynamically changing fields (like for dynamic select choices), dynamically added fields, etc. Probably would try to turn those into helpers as well. You pass in fields, you get an event that says dynamic changes are required, you use helpers to modify the fields, then you supply new fields.


bryanhelmig:
So much merging today!

Bryan

PR #16, opened 2015-01-20
Need default values.

jdeal:
If you do something like this:

```js
var Form = React.createFactory(Formatic);
...
React.render({
  fields: [
    {
      type: 'text',
      key: 'title',
      default: 'Untitled'
    },
    {
      type: 'text',
      key: 'author',
      default: 'Unknown'
    }
  ],
  value: {},
  onChange: function (newValue) {
    console.log(newValue);
  }
}, document.body)
```

Should we immediately see `{title: "Untitled", author: "Unknown"}` logged to the console?

Or, should that only occur after the first change? So if I change "Unknown" to "J", then I see `{title: "Untitled", author: "J"}`.

I'm leaning toward the latter, given an immediate `onChange` would imply that the user had changed something, when that's not really what happened. We can later add some other event like `onDefault` for catching the case that default values are created, if necessary.

Right now, neither of these is occurring. Default values essentially have no effect, because changes only bubble up when fields are changed. So changing "Unknown" to "J" would just give you `{author: "J"}`. Worse than no effect, they are actually misleading. They imply that a field has a value, but until the field is touched, it really doesn't have that value.


jdeal:
Another nuance here.

With this example:

```js
var Form = React.createFactory(Formatic);
...
React.render({
  fields: [
    {
      type: 'text',
      key: 'title',
      default: 'Untitled'
    },
    {
      type: 'text',
      key: 'author'
    }
  ],
  value: {},
  onChange: function (newValue) {
    console.log(newValue);
  }
}, document.body)
```

Where I have no default for author, is the default implied to be a blank string (`""`)? In other words, if I change the title to 'Foo', do I get back `{title: 'Foo', author: ''}`, or do I get back `{title: 'Foo' }` until an explicit edit to author is made?

A further wrinkle here is if I set the author but then erase it by clearing the field, what is the value of the field then? Is it a blank string, or is the key/value removed?

jdeal:
More nuances/thoughts:

If a user doesn't manipulate a form at all but is happy with default values, then without some event, there is no way to retrieve the value. With the new simplified formatic, there is no value other than what is passed in and what is sent back in the onChange callback. So if we want the initial value, we either have to send that out in an onChange event or some other event, like onDefault mentioned above. This is probably a bit of an edge case, but it's still a problem none-the-less. If formatic ever displays a different value than what was passed in, there's potential for confusion.

We can't clear fields when a user blanks out a field, because having no value is exactly what should signal the default value in the first place. (I think there's actually another problem currently where formatic treats empty strings as non-existent values.) As far as formatic is concerned, there's no "beginning" state. It's just what's passed in now. So we have to look at the data every time and work from that.

jdeal:
Okay, here's where I landed so far:

If you pass in a non-empty value (empty being undefined, empty string, null, or object with no keys), that's the value you will get back in an `onChange`. Formatic assumes you meant for `{foo: 'bar'}` to be your initial value, and it will not give you back other default values for other keys. It will only change/append those values.

If you pass in an empty value, it will assume you want the defaults, and it will "inflate" the fields to get an initial value. You will not receive these values in an `onChange` until something is edited by the user. At that point, you'll get all the default values along with the new changes.

All this behavior is now in the config (it used to be locked in the root component) split across several helper functions, so it's possible to tweak it with a plugin.

If you want to grab the intiail default values so that a user could save the defaults, then you can manually call this before passing the value to the root Formatic component:

```js
var value = config.createRootValue(props);
```

`props` in this case is the same set of props you would be sending to the root Formatic component, so you'll know you're getting exactly the same default values that the components get. For example:

```js
var value = config.createRootValue({fields: myFields, value: {}});
```

Would give you an initial value that would match what the user sees, assuming you pass the same props to the root component, obviously. If you later add some fields, and you want default values for those fields as well, you could do this:

```js
var value = config.createRootValue({fields: myFields, value: myCurrentValue});
```

This would merge in the new defaults to your current value, so you have a value that matches what the user sees. That "merge" is a little spooky, but as long as components correctly rely on the config to get values, the merge should result in values that are exactly the same as what the component displays.

jdeal:
Note that all fields have implicit default values now as well. So the default value for a text field is a blank string. The logic here is:

1. This matches the expectation from a standard HTML form. The field implies a value, even if it's a blank value.
2. This makes the value of a field consistent no matter if it was ever edited or not. If we say that cleared fields should be removed, then what stops the default value from showing up again? If I backspace and remove a value, should the default value really pop again? That would be rather strange.

Apps that want to remove blank fields thus have to remove the blank fields manually. (Although we could add a helper or possibly even a setting for this.)

Another option is to not have implicit default values. Apps would be required to do the reverse, to ask for implicit defaults. This might complicate the logic a bit, because those implicit defaults are currently in place to make components simpler (so they can always expect a value). So there might be some conditional execution paths there.

jdeal:
Comical results of using default value if current value is a blank string:

blank-default

(Fixed above.)

jdeal:
Going ahead and merging this for now. Can adjust later if there are good reasons to do so.

PR #17, opened 2015-01-21
Going to clean up comments.

jdeal:
Some comments are out of whack from refactor. Need to clean those up.


jdeal:
Lots more could be done, but at least readme and comments are up-to-date and corrected with refactor and also added a few more comments.

lib/default-config.js
```diff
@@ -702,26 +750,28 @@ module.exports = {
     return field.itemFields;
   },
 
-  fieldIsSingleLine: function (field) {
-    return field.isSingleLine || field.is_single_line || field.type === 'unicode' || field.type === 'Unicode';
-  },
+  fieldIsSingleLine: utils.delegateTo('fieldTemplateIsSingleLine'),
```
bryanhelmig: Like the move to utils namespace here.

PR #18, opened 2015-01-22
Need working tests again.

jdeal:
Tests have been broken for too long. Going to get those working again.


jdeal:
Merging what I got so far, as some tests are better than no tests! Still need to do something for the pretty-text field.

From now on, since the api should be pretty stable, I'm going to try to work test first. @bcooksey @bryanhelmig let me know if you have any questions about how to write tests.

PR #20, opened 2015-02-10
Create some docs.

jdeal:
Going to get some minimal docs out there.

- [x] Gulpify groc. Currently, groc publishes annotated source to http://zapier.github.io/formatic/index.html. We need other docs at the root, so gulp needs to move to a subpath.
- [x] Add gulp task for publishing annotated source. (To replace groc's default publishing.)
- [x] Create a homepage for formatic.
- [x] Make a formatic sandbox page. Allow editing of examples to see resulting forms.
- [ ] Make some examples to show different field types, etc.


jdeal:
Merging this even though I still need to add some more examples. Will start another PR for those.

PR #21, opened 2015-02-15
Document all field types

jdeal:
All the field types are now documented.


bryanhelmig:
Excited to see this upgraded into zapier/zapier - I ran into a few of the bootstrap bugs in my HTTP tool.

jdeal:
react-bootstrap seems pretty solid, though admittedly, I was just doing a bunch of copy pasta.

It definitely demonstrates the clear need for JSX though. Canonical bootstrap is HTML, and JSX is a much better analog for that.

bryanhelmig:
Seems like the writing is on the wall WRT Coffeescript... :-)

PR #23, opened 2015-03-04
Clean up gulp

jdeal:
The branch is called remove-gulp because I was trying to do that. However, I ended up keeping gulp around for a few reasons:

1. It's more performant and/or easier to manage task dependencies. Using just `npm run` commands, there's no easy way to have several commands that all depend on another command. gulp does a pretty good job of managing that and making it parallel when possible.
2. I can comment my gulp tasks. You can't comment a package.json file, so there's no way to comment the commands when they get complex. With gulp, things are self-documenting most times anyway, because of the list of deps.
3. It's more naturally portable. I used shelljs for portability, but I ended up having to write tiny scripts just to copy files, which seemed kind of crazy.
4. I have fine-grained control when a task is finished. For some things like watchify, I want to consider the task finished when the initial file is built. With command line scripts, my choice is either way till the script is finished or let it run forever. There's no middle ground. Without this fine-grained control, it's easy to end up with things stomping on each other.

The primary disadvantage to gulp is that when things are wrapped in plugins, we end up with stale versions of things. As such, from now on, I am going to avoid plugins where possible.

Another disadvantage arose with gulp itself having some nested out-of-date deps which conflicted with the goal to use 6to5 (now babel). This happened when I needed to pre-compile some react-based docs. To get out of this problem, I put that stuff in a script and delegate to it command-line style from gulp. I think this is also a good goal forward. Rather than having gulp do to much, make smaller scripts that do work and call them as a process, so the code is separately executing.

In general, I tried to clean up the gulp tasks also. Docs are now watched as well.


jdeal:
Oh, also, I want to use `npm run *` commands as the entry points. This way, you don't actually have to install gulp. (Need to change the README to reflect that.)

PR #25, opened 2015-03-06
Pretty text with code mirror

stevemolitor:
Implements the PrettyText component using CodeMirror.  For efficiency CodeMirror is swapped in and out as the user mouses in and out of the control.  When outside the control a less expensive read-only view is displayed.

Look at ```/lib/components/pretty-text2.js``` and files required from it.  We'll rename it to pretty-text when done.

You can see it in action by firing up the demo:  

```bash
git clone git@github.com:zapier/formatic.git
cd formatic
# or
git checkout pretty-text-with-code-mirror
git pull
# then...
npm install
npm run live
```

Point your browser to `localhost:3000/index.html`.  The first component will be the pretty-text2 component.

## Features TODO 
- Add support for adding an image to a pill box, 
- Make it work in mobile. For example maybe we switch to codemirror in onTouch.

## Code TODO
- Remove hard-coded Bootstrap classes, make the BS stuff a plugin
- Implement ChoicesDropdown without Bootstrap (I don't think we want to force that dependency - just steal the code from the old PrettyText component)
- Normalize some of the names in the code - tag vs. key vs. choice etc. - pick one!


mikeknoop:
@stevemolitor will check out soon. I'd close this and re-open a new PR though to `develop` branch first.

We only merge `develop` in `master` right before a deploy so that `master` is always safe to fall back to.

Might have to play w/ the history in case you branched off of `master` and not `develop`.

mikeknoop:
Is the `demo` directory intended to be committed? Lot's of files in there!

stevemolitor:
@mikeknoop On formatic we branch off of master into feature branches for development.  Releases will have their own branches or tags. We don't have a develop branch. @jdeal correct me if I'm wrong.

All of the files in the demo directory are from the bower install of codemirror. We put external dependencies like underscore and codemirror in the repo but not the formation distribution (formatic-dev/min.js) to make running the demo easy.  However the codemirror bower install does have a lot of files most of which we don't use.  So we could slim that down.  Anyway for now just ignore everything under ```demo/bower_components```.

mikeknoop:
:+1: (didn't realize this was the `formatic` repo w/ my first, now deleted, comment).

bryanhelmig:
We should slap a invisible textarea to catch tabs:

track-tabs

PR #30, opened 2015-04-28
Remove this from config

jdeal:
@stevemolitor This tweaks the plugin system so that `this` is no longer necessary for plugins. Whereas a plugin used to have to be:

```js
var upperHumanizePlugin = function (config) {
  
  humanize = config.humanize;
  
  config.humanize = function (key) {
    return humanize.call(this, key).toUpperCase();
  }

  config.fieldLabel = function (field) {
    return this.humanize(field.label || field.key);
  }
};
```

It can now be:

```js
var upperHumanizePlugin = function (config) {
  
  humanize = config.humanize;
  
  config.humanize = function (key) {
    return humanize(key).toUpperCase();
  }

  config.fieldLabel = function (field) {
    return config.humanize(field.label || field.key);
  }
};
```

This is because all plugins get passed in the same `config` object. so their methods always have access to that object. I've also added the ability to define the plugin like this:

```js
var upperHumanizePlugin = function (config) {
  
  humanize = config.humanize;
  
  return {
    humanize: function (key) {
      return humanize(key).toUpperCase();
    },
    fieldLabel: function (field) {
      return config.humanize(field.label || field.key);
    }
  };
};
```

Which I think I prefer.


jdeal:
@stevemolitor This is definitely still open to comments, but I'm merging as it seems like a decent improvement in terms of reducing complexity.

PR #32, opened 2015-04-30
Changes wording to make Shared Zap creator more clear

craiglabenz:
As per @benpeter's suggestion
@jdeal -- is this an acceptable text change? (Are there any places where it would actually be *incorrect*?)


bryanhelmig:
Funny enough @stevemolitor is working on making this an element/helper we can override. Gonna close for now!

Also in general - the Formatic repo is not the place to inject Zapier specific phrasing.

PR #33, opened 2015-05-08
Pretty select

stevemolitor:
@jdeal @bryanhelmig 

WIP. Created new PrettySelect field that displays fancy drop downs with key - value, etc. It's a psuedo-select, not native browser select. Uses <input> and the choices helper for the drop down.  Eventually need to style the input to make it look like a drop down with an arrow image on the right, etc etc.


stevemolitor:
@jdeal The styling is lame but I think it's ready for trying out in zapier.  Give it a once over if you have a chance.

PR #37, opened 2015-05-11
Only render choices-container when open

stevemolitor:
@bryanhelmig @jdeal 

There was a display bug in zapier when choices are closed:  a  pixel displayed from the `.choices-container` border style.  A fix was to add `.choices-conainer-closed` when closed.  Unfortunately this breaks the CSS transition - the second time you open a choices container it displays blank because the CSS transition even never gets fired.  

My solution here was simply to only display the `choices-container` wrapper div when choices are open, and otherwise display an empty div.


lib/components/helpers/choices.js
```diff
@@ -109,52 +109,51 @@ module.exports = React.createClass({
...
-      )
-    );
+            }.bind(this))
+          )
+        )
+      );
+    }
+
+    // not open
+    return R.div(null);
```
bryanhelmig: Why not `return null;`? Does that break React?

PR #40, opened 2015-05-12
pretty-boolean, fix pretty-select choices issue

stevemolitor:


bryanhelmig:
What is up with this extra PR? Isn't this already in master? Can we close this?

PR #41, opened 2015-05-13
Support custom actions in pretty-select choices

stevemolitor:


bcooksey:
@stevemolitor Is this a stale PR? It looks like a version of this code is already in master. If so, let's close this one

PR #47, opened 2015-06-03
Immutable

jdeal:
@stevemolitor @bryanhelmig Here's a taste of what I was toying with re formatic changes.

Probably messy to see what I'm doing but if you look here:

https://github.com/zapier/formatic/blob/f59f9cb3bdb0643c71349df0421954fa6398b6a5/demo/index.html#L44-L96

(You can also run the demo if you like--this much works.)

That gives you an idea of what I'm going for. Formatic would just be a bunch of components with some components supporting an alternate JSON `fields` prop. Note the two ways shown of creating a form, one with JSX, another with JSON.

Plugins would simply be providing different components which could then easily delegate back to the default components. A toy example where I bold the Label is shown.

Other goals would be:
- Assume immutable data by default so we can do quick `shouldComponentUpdate` comparisons.
- Get rid of the contextual wrapping of fields so as not to defeat immutable checks.
- Make lookups for dropdowns a first-class citizen (simple metadata lookup props passed in).
- Other than the component override/wrapping described, reduce or possibly eliminate all other extension points.


bryanhelmig:
Wow - the JSX syntax looks impressive and mighty friendly!

I'd have to spend a bit of time trying to wrangle a usable sample from the code to have much of substance to say - but definitely interested in ripping out the customizable mutators (or field wrappers, as you said).

stevemolitor:
Looks awesome.  Would this work:

<Fields.Fields config={config} onChange={(value) => console.log('JSX:',
value)} defaultValue={{}}>
  <Fields.String fieldKey="greeting" label="Greeting" helpText="Say hello
to your momma."/>
*  <p>Stuff in between fields</p>*
  <Fields.String fieldKey="name" label="Name" helpText="The name your momma
gave you."/>
</Fields.Fields>

jdeal:
Yes, that could work. And I just tried it, and it did work.

Only caveat is that `Fields.Fields` is passing along `config` and `onChange` to its children, because it assumes they want it. It might eventually pass other properties like `meta` or whatever. In certain rare edges, those could collide with child properties. We _might_ be able to have the parent component inspect the children to know if they're "formatic" components. Otherwise, might have to have something like this for those edges:

```
<NotFormatic><p>Stuff in between fields</p></NotFormatic>
```

stevemolitor:
Cool!

Also if an 'in between' component uses propTypes to specify allowed props,
then it'd spew warnings to the console when `config` and `onChange` aren't
in its propTypes.  So yeah, <NotFormatic> might help for edge cases like
this.  Not a biggie though just wondering.

PR #52, opened 2015-07-02
Custom pretty select input

bcooksey:
We'd like to have the PrettySelect's input be customizable. Currently is hard-coded to an input that is either editable or read-only.

Additionally, we are going to see how it works with having PrettyText be the default text box.


bcooksey:
@jdeal some further context. For part of this (or a future PR), I'm going to want to include the ability to insert fields when typing a custom value into a pretty select. The styling is different on those, however, because we'll want to hide the menu except when click the new menu option. I'm not sure if that informs which kind of split is better for the PrettyTextHelper (put choices in the field wrapper or choices in the helper itself), but it's another something to consider.

PR #53, opened 2015-07-06
add password field

stevemolitor:
@jdeal


lib/components/fields/password.js
```diff
@@ -0,0 +1,43 @@
...
+
+'use strict';
+
+var React = require('react/addons');
+var R = React.DOM;
+var cx = require('classnames');
+
+module.exports = React.createClass({
+
+  displayName: 'SingleLineString',
```
jdeal: displayName should be `Password`

PR #56, opened 2015-07-16
add grouped fields component

stevemolitor:


build/lib/components/fields/grouped-fields.js
```diff
@@ -0,0 +1,108 @@
...
+    var legend;
+    var className = cx(this.props.classes);
+
+    if (groupLabel) {
+      legend = React.createElement(
+        'legend',
+        null,
+        groupLabel
+      );
+      className += ' fields-group';
```
stevemolitor: hacky - use `cx `

jdeal:
@stevemolitor How is the "grouped-fields" type different from the "fields" type?

PR #57, opened 2015-07-16
add keys to grouped fields

stevemolitor:


build/lib/components/fields/grouped-fields.js
```diff
@@ -79,12 +79,12 @@ module.exports = React.createClass({
         null,
         groupLabel
       );
-      className += ' fields-group';
+      className += ' child-fields-group';
```
stevemolitor: still hacky - use `cx `

PR #63, opened 2016-01-13
Add AssocList field to (eventually) replace Object field

stevemolitor:
The Object field is buggy and brittle because its actual state (what the user sees) doesn't always map to its logic field value (an object). While editing the user can temporarily see rows with duplicate or blank keys, but we map it back to an object, which can't have duplicate keys. To avoid UI weirdness (like rows poofing out of existence when a key becomes a duplicate), the object field had lots of complicated logic involving pollution of the field value with secret values, and other fun stuff like that. But we still have bugs with fields poofing out of existence, and field keys becoming blank.

This PR gets rid of all that. The Object field is replaced with a new AssocList field. It displays an array of key / value pairs, with duplicates allowed: 

```js
  // field def:
  {
    label: 'Associative List', type: 'assoc-list', key: 'assoc-list',
    itemFields: [{
      type: 'pretty-text',
      hideLabel: true
    }],
    default: [{key: 'key1', value: 'value1'}, {key: 'key2', value: 'value2'}]
  },
```

What you see is what you get - the visible state matches the field value. There's no complicated logic. For efficiency it uses React `update` to merge in and change rows.

Eventually maybe we can stop using the object field, and coalesce field values to and from associative arrays into objects when loading and saving (which would be efficient). But in the interim the old object field has been gutted to use AssocList under the hood, converting on onChange (which is not so efficient).

It's late so I'm gonna proof read this tomorrow but I thought I'd throw it out there for feedback before bed.

I changed the `object-item-*` helpers for consistency, which unfortunately means I'll have to change the `object-item-*` css class names upon merging to zapier. ;( I checked, there's not too many though.

@jdeal @bryanhelmig


stevemolitor:
Looks like I've got some failing tests. Will fix tomorrow.

stevemolitor:
The gutted object field doesn't do componentWillRecieve props - probably ok for our usage?

jdeal:
@stevemolitor I always meant to return to this some day as it's hella goofy, but the intention of all that goofiness was to:

1. Keep fields from moving on you as you change keys. It's especially weird if you're typing, and that causes the key order to change, and then your key input moves off the screen. This is why I retained the state of the keys: so that new keys were guaranteed to be appended, never inserted above you.
2. Make sure your focus never moves to a different key while you're typing, which can happen if you're not careful with React keys.

I've only glanced at your code, but I'm not seeing use of a `key` prop at all in your AssocList component, so I'm curious if there's any goofiness there. By not using a key, it's position-based (I think), which, I'm not following how you can guarantee that the position for a key will always remain the same. If anything outside the component mucks with the object, things could get weird.

I'm all for simplifying the code, and conceptually, I like the idea: having an object component that takes objects and converts them to key/value pairs to be handed off to a component that takes key/value pairs. I just want you to be aware of potential edges.

stevemolitor:
@jdeal Yes great catch about keys. I added code to maintain persistent artificial keys here: 8f3c0d2. I followed the pattern in the array field. (It's a bit simpler in this case because we don't support moving rows around in this component, but we do support deleting a row which shuffles rows down.)

I think there were two sources of complexity in the origin Object field component:
1. Maintaining artificial keys
2. Allowing user to enter duplicate keys when the underlying data structure (object) does not allow duplicate keys

\#1 is necessary complexity as you point out. I think having a component that takes key / value pairs like this greatly reduces the complexity of #2, so still a win.

Slightly off topic, but I wish 'saving' zap changes in flow was more explicit. We could still persist 'unsaved' changes somewhere (local storage, a special spot on server).

jdeal:
I have a feeling there are still some edges that aren't covered. Try this in the demo:

1. Hook up `setValue` to `window` so you can change the value of an object field via the console.
2. Set the value of an object field to `{a: 'foo', b: 'bar'}`.
3. Change the value of the field to `{b: 'bar', a: 'foo'}`, via `setValue`.

Because you're using indexes as keys, I think you'll end up with `{a: 'bar', b: 'foo'}` in your AssocList component. Then if you go on to tweak a value, say change 'foo' to 'fooo'. You'll end up with:

`{a: 'bar', b: 'fooo'}`

Essentially, I think you're trying to treat objects as arrays, and they're not. With the array component, if you switch `['foo', 'bar']` to `['bar', 'foo']`, then yes, those should swap places, but with objects, it's a bit more nuanced.

stevemolitor:
@jdeal More changes. I tested your scenarios above and they work with these changes.

- Highlight all rows with duplicate keys:

dups

- Do not fire onChange events when there are duplicate keys
- Ignore changes from the outside (`componentWillReceiveProps`) when there are dups
- Add unit tests for these scenarios

I think it's looking fairly solid now.

PR #64, opened 2016-05-04
Allow sorting of choices

stevenhauser:
Hold onto your butt @jdeal, this is a big one.

Also what's the release for this like? My guess is…

```
npm run build
npm run patch-release
```

Though this is more of a minor than a patch to me, but I dunno if that's much of an issue here.

### DIFF
--- a/lib/components/helpers/choices.js
+++ b/lib/components/helpers/choices.js
@@ -237,6 +237,8 @@ module.exports = React.createClass({
       });
     }
 
+    choices = config.sortChoices(choices, this.state.searchString);
+
     if (!props.isAccordion) {
       return {
         choices,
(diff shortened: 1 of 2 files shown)


jdeal:
I've been committing the build to the repo. Don't ask me why I did it that way. Probably hurried laziness.

So you'll have to commit and push between those two steps for now. Unless you want to fix that. (Delete `/lib` from github, add `/lib` to `.gitignore`, add a prepublish step.) But certainly no requirement to do that for this PR.

jdeal:
And yes, technically, this would be a minor bump, but it's a really minor minor bump, so I'll leave it to you. 😄 So far, formatic hasn't really followed semver very well.

stevenhauser:
Okeydokey that's all fine, I'll go the path of least resistance for now since I don't think there's a major reason to diverge.

PR #65, opened 2016-07-06
add `onTouchStart` to pretty-text-input

codekirei:
Tested this in Chrome's device simulator, and with this change, clicking an input box to the right of a pretty pill correctly places the cursor directly to the right of the pill as expected.

With change:

Previously:

edit: this will need an `npm version patch` &mdash; not sure if I should add that to this PR or do it separately


codekirei:
patching separately from this branch.

PR #66, opened 2016-08-22
Fix how placeholders work in PrettyTextInput

stevenhauser:
- Use `placeholder` option instead of always using `value`
- Simplify `hasPlaceholder` to `!this.state.value` rather than checking
`hasChanged`, etc. This doesn't appear to have any negative side
effects.


stevenhauser:
@jdeal wanna put a quick set of 👀  on this?

jdeal:
@stevenhauser Weird that it was done this way originally. Can only guess it was just a misunderstanding of the `placeholder` property. This seems straightforward enough, so LGTM.

PR #67, opened 2017-01-19
Add ability to disable search box in pretty-select

neilff:
- Adds `disableSearch` prop to `pretty-select` field that when set to `true` will disable the search box, useful for short dropdowns

### DIFF
--- a/lib/default-config.js
+++ b/lib/default-config.js
@@ -878,7 +878,7 @@ module.exports = function (config) {
     },
 
     fieldHasSearch: function (field) {
-      return field.disableSearch !== true ? true : false;
+      return _.isUndefined(field.hasSearch) ? true : field.hasSearch;
     },
 
     // Determine if value for this field is not a leaf value.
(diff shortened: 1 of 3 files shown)


lib/default-config.js
```diff
@@ -877,6 +877,10 @@ module.exports = function (config) {
       return field.required ? true : false;
     },
 
+    fieldHasSearch: function (field) {
+      return field.disableSearch !== true ? true : false;
```
jdeal: @neilff Can we reverse the logic like this:

```js
return _.isUndefined(field.enableSearch) ? true : false;
```

or `hasSearch` would be okay too. (Formatic has a mess of conventions with booleans right now, so you'll notice the boolean prefixes are not consistent. 😞 Guess we might as well move toward boolean prefixes though.)

I wish I had done this in the first place, and then I'd force us to just enable search when we want it. I don't want to break everything, so hence keeping the default as `true`.

neilff: Sure that makes sense to me, I'll apply those changes

PR #68, opened 2017-02-06
Add ability to customize fieldset form wrapper

neilff:
@jdeal can you provide feedback on this PR? 

It seemed a bit odd that I had to wrap the `children` in a `<div>`, is there a way to avoid this? passing them as separate args was not working as expected.

I plan to remove the example plugin when this is confirmed 🆗

### DIFF
--- a/demo/index.js
+++ b/demo/index.js
@@ -12,6 +12,7 @@ var plugin = function (config) {
 
   var initField = config.initField;
   var createElement_PrettyTag = config.createElement_PrettyTag;
+  var createFieldset = config.createElement_Fieldset;
 
   config.createElement_PrettyTag = function (props, children) {
     //var choice = _.find(this.props.replaceChoices, (c) => c.value === tag);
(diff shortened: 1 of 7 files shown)


demo/index.js
```diff
@@ -47,6 +47,16 @@ var plugin = function (config) {
 
     isRemovalOfLastAssocListItemAllowed(/*field*/) {
       return false;
+    },
+
+    createElement_Fieldset: function (props, ...children) {
```
jdeal: I wonder if maybe you want to make this conditional for a custom property on `props.field`, just to remove noise in the demo? Otherwise, the next developer may be confused why "I'm a custom fieldset!" is showing up everywhere.

neilff: dah, good point, I'll take care of that

neilff: @jdeal can you take a look at my updated example? I had to do a bit more work than I had expected, tell me if it makes sense :/

jdeal: Yep, that's totally fine @neilff! I forgot, the `fieldset` helper doesn't correspond to a field that's passed in. Should have told you that you could just do a `{type: 'fields', fields: [...]}` wrapper around some fields. But this works too!

PR #71, opened 2017-03-13
Remove falsey values from choices in normalizeChoices

stevenhauser:
It's possible that `null` values are sent for instance, and
`normalizeChoices` throws when that happens because we try to access
`label` off of `null` a little bit below this LOC. Ideally we would
always receive good `choices` but we need to handle cases where choices
are invalid.

### DIFF
--- a/lib/default-config.js
+++ b/lib/default-config.js
@@ -1050,7 +1050,7 @@ module.exports = function (config) {
       choices = choices.slice(0);
 
       // Array of choice arrays should be flattened.
-      choices = _.flatten(choices);
+      choices = _.compact(_.flatten(choices));
 
       choices.forEach(function (choice, i) {
         // Convert any string choices to objects with `value` and `label`
--- a/__tests__/undash.js
+++ b/__tests__/undash.js
@@ -10,6 +10,17 @@ describe('utils', () => {
     expect(_.flatten([['foo', 'bar'], 'baz'])).toEqual(['foo', 'bar', 'baz']);
   });
 
+  it('should compact', () => {
+
+    expect(_.compact([])).toEqual([]);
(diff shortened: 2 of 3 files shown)


stevenhauser:
@jdeal can you 👀 this and let me know if you see any problems with it? I could see doing a `filter` as well and _only_ accepting strings or objects with the correct format, though that would fail silently on more cases which wouldn't be great. Or instead of doing a `filter` we could throw an explicit error.

lib/default-config.js
```diff
@@ -1050,7 +1050,7 @@ module.exports = function (config) {
       choices = choices.slice(0);
 
       // Array of choice arrays should be flattened.
-      choices = _.flatten(choices);
+      choices = _.compact(_.flatten(choices));
```
jdeal: Just to keep things interesting, formatic doesn't use lodash or underscore but rather its own undash lib: https://github.com/zapier/formatic/blob/28cb22cc9d970c8f9da5efe2e725799ca8e07b78/lib/default-config.js#L14. :-)

So you'll need to add `_.compact` to that.

This was done to keep dependencies very small. Probably should have not called it `_` though and just called it 'utils' or something.

stevenhauser: Ahhh yikes, yeah I didn't run it in the demo, I `npm link`ed it. I'll be sure to use the demo moving forward.

stevenhauser: Ok @jdeal I've added `_.compact`. Good to merge?

stevenhauser: Actually I'm going to write a couple quick tests…

stevenhauser: Ok tests written, ready for rereview @jdeal

PR #73, opened 2017-04-24
Upgrade to React 15.5.0

neilff:
- Upgrades to React and React-DOM v15.5.0
- Upgrades `react-hot-loader` to v1.3.1
- Adds `prop-types` and `create-react-class` as dev. dependencies
- Applies [`jscodeshift` to use `create-react-class`](https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#migrating-from-react.createclass)
- Applies [`jscodeshift` to use `prop-types`](https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#migrating-from-react.proptypes)


ngryman:
ref #85

PR #75, opened 2017-05-02
Fix choice with action jsx

stevemolitor:
This commit #http://localhost:8000/app/editor/2078/nodes/2079/fields inadvertently changed things when converting to JSX. What was two sibling `span` elements got nested. This PR reverts that.


neilff:
ah, good find @stevemolitor. This change looks good to me

stevemolitor:
No worries @neilff!

PR #79, opened 2017-06-01
Add updates for reorder-items-in-zap-editor

efhjones:
To be shipped with https://github.com/zapier/zapier/pull/12121

Just a few updates to the formatic code for reordering the items in the zap editor -- modifies the code a bit to show the `move-item-back` and `move-item-forward` arrows only when necessary, and adds a class to the first div child of `pretty-text-wrapper` as that needed a slight tweak for styling.

### DIFF
--- a/lib/components/helpers/array-item-control.js
+++ b/lib/components/helpers/array-item-control.js
@@ -44,13 +44,13 @@ export default React.createClass({
       readOnly: isLastItem && !config.isRemovalOfLastArrayItemAllowed(field)
     });
 
-    const moveItemForward = field.fieldIndex !== field.parent.value.length - 1 ?
+    const moveItemForward = this.props.index < (this.props.numItems - 1) ?
       config.createElement('move-item-forward', {
         field: field, onClick: this.onMoveForward
       }) :
       null;
 
-    const moveItemBack = field.fieldIndex !== 0 ?
+    const moveItemBack = this.props.index > 0 ?
       config.createElement('move-item-back', {
         field: field, onClick: this.onMoveBack
       }) :
--- a/lib/components/helpers/array-item-control.js
+++ b/lib/components/helpers/array-item-control.js
@@ -45,13 +45,11 @@ export default React.createClass({
     });
 
     const moveItemForward = this.props.index < (this.props.numItems - 1) ?
-      config.createElement('move-item-forward', { field: field, onClick: this.onMoveForward
-      }) :
+      config.createElement('move-item-forward', { field: field, onClick: this.onMoveForward }) :
       null;
 
     const moveItemBack = this.props.index > 0 ?
-      config.createElement('move-item-back', { field: field, onClick: this.onMoveBack
-      }) :
+      config.createElement('move-item-back', { field: field, onClick: this.onMoveBack }) :
       null;
 
     return (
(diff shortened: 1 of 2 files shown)


lib/components/helpers/pretty-text-input.js
```diff
@@ -283,7 +283,7 @@ export default React.createClass({
     return (
       <div onKeyDown={this.onKeyDown} className={cx({'pretty-text-wrapper': true, 'choices-open': this.state.isChoicesOpen})}
            onMouseEnter={this.switchToCodeMirror} onTouchStart={this.switchToCodeMirror}>
-        <div onClick={this.onFocusClick}>
+        <div className="sub-pretty-text-wrapper" onClick={this.onFocusClick}>
```
jdeal: That's a pretty funny name. Maybe `pretty-text-click-wrapper`? Still a funny name, but maybe a bit more descriptive. BTW, reason for this `div` here: https://github.com/zapier/formatic/commit/cbf9680a56e94fa5c3251c4d23dc95bfd2d496be#commitcomment-20264080.

lib/components/helpers/array-item-control.js
```diff
@@ -44,12 +44,16 @@ export default React.createClass({
       readOnly: isLastItem && !config.isRemovalOfLastArrayItemAllowed(field)
     });
 
-    const moveItemForward = this.props.index < (this.props.numItems - 1) ?
-      config.createElement('move-item-forward', { field: field, onClick: this.onMoveForward }) :
+    const moveItemForward = field.fieldIndex !== field.parent.value.length - 1 ?
```
jdeal: Shouldn't `this.props.numItems` be the same as `field.parent.value.length`? I'd like to keep the parent responsible for that, rather than this control having to peek at the parent value.

I'm not quite following how this changes the logic from what it was before: only showing the forward button if it's not the last item.

lib/components/helpers/array-item-control.js
```diff
@@ -44,12 +44,16 @@ export default React.createClass({
...
+      config.createElement('move-item-forward', {
+        field: field, onClick: this.onMoveForward
+      }) :
       null;
 
-    const moveItemBack = this.props.index > 0 ?
-      config.createElement('move-item-back', { field: field, onClick: this.onMoveBack }) :
+    const moveItemBack = field.fieldIndex !== 0 ?
```
jdeal: Same question here: `field.fieldIndex` should be the same as `this.props.index` so I'm not seeing how this changes anything. Maybe I'm mistaken though, or maybe there's a bug with `index`?

lib/components/helpers/pretty-text-input.js
efhjones: @jdeal Ah ok thanks! Yep, can change.

lib/components/helpers/array-item-control.js
efhjones: Thanks for taking a look @jdeal ! Most likely will just switch these back. I remember hunting for a bug last week-- this probably got changed as a result of that (though the bug was elsewhere) and then left in when it worked 😬  but I switched it back and didn't see any problems. Sorry for the confusion! Will push up the change.

lib/components/helpers/pretty-text-input.js
efhjones: Ok cool, thanks for pointing that out! Found the bug and fixed in https://github.com/zapier/zapier/pull/12121

PR #80, opened 2017-06-06
Treat 0 as false when coercing a value to a boolean

stevenhauser:
@jdeal see any issues with this?


jdeal:
@stevenhauser This seems reasonable in that it's a better match for JS semantics.

PR #82, opened 2017-08-17
Open choices on Focus for Insert Fields

efhjones:
Retreat August 2017 project: Insert Fields UX

This PR auto-opens the insert fields dropdown when focused into the text box. For use alongside https://github.com/zapier/zapier/pull/12880

### DIFF
--- a/lib/components/fields/pretty-text2.js
+++ b/lib/components/fields/pretty-text2.js
@@ -40,6 +40,7 @@ export default React.createClass({
       field: this.props.field,
       value: this.props.field.value,
       isAccordion: this.props.field.isAccordion,
+      shouldOpenChoicesOnFocus: this.props.field.shouldOpenChoicesOnFocus,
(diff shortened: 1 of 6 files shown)


stevemolitor:
This is pretty much all retreat hacks - we can't release this as is. A few problems of the top of my head:

1. We changed the behavior of _all_ pretty-text2 components. So if you use pretty-text2 in your site, this will change _all_ of them to pop open the dropdown, including when pretty-text2 is used a a sub component of another field type, like adding a custom field value.
2. Keyboard navigation doesn't work with this new feature - tabbing in and out doesn't pop open or close the dropdown properly. 
3. Not really sure if we hacked it into the best spot in the code.

I'm a little worried about what it would take to make this real - we'd have to handle the keyboard navigation and such, adding yet more code to make a non-native control feel native-ish.

efhjones:
Going to close this for now since this was experimental.

PR #83, opened 2017-09-27
Recognize replaceChoices changes and allow tagLabel

efhjones:
This PR adds a check for `replaceChoices` changes to allow those to drive an update for codemirror. 

This PR also introduces support for `tagLabel`, which allows the user to determine the specific label for use in the tag without affecting dropdowns or other values.

### DIFF
--- a/lib/components/helpers/pretty-text-input.js
+++ b/lib/components/helpers/pretty-text-input.js
@@ -42,7 +42,14 @@ export default React.createClass({
   componentDidUpdate: function(prevProps, prevState) {
     const wasPreviewing = prevProps.field.meta.isPreviewing;
     const isPreviewing = this.props.field.meta.isPreviewing;
-    if (prevState.codeMirrorMode !== this.state.codeMirrorMode || wasPreviewing !== isPreviewing) {
+    if (wasPreviewing !== isPreviewing) {
+      this.setState({ codeMirrorMode: false}, () => {
+        this.createEditor();
+      });
+    }
+    if (
+      prevState.codeMirrorMode !== this.state.codeMirrorMode
+    ) {
       // Changed from code mirror mode to read only mode or vice versa,
       // so setup the other editor.
       this.createEditor();
--- a/lib/components/helpers/pretty-text-input.js
+++ b/lib/components/helpers/pretty-text-input.js
@@ -40,15 +40,10 @@ export default React.createClass({
   },
 
   componentDidUpdate: function(prevProps, prevState) {
-    const wasPreviewing = prevProps.field.meta.isPreviewing;
-    const isPreviewing = this.props.field.meta.isPreviewing;
-    if (wasPreviewing !== isPreviewing) {
-      this.setState({ codeMirrorMode: false}, () => {
-        this.createEditor();
-      });
-    }
+    const replaceChoicesHaveChanged = prevProps.replaceChoices !== this.props.replaceChoices;
     if (
-      prevState.codeMirrorMode !== this.state.codeMirrorMode
+      prevState.codeMirrorMode !== this.state.codeMirrorMode ||
+      replaceChoicesHaveChanged
     ) {
       // Changed from code mirror mode to read only mode or vice versa,
       // so setup the other editor.
--- a/lib/components/helpers/pretty-text-input.js
+++ b/lib/components/helpers/pretty-text-input.js
@@ -40,10 +40,9 @@ export default React.createClass({
   },
 
   componentDidUpdate: function(prevProps, prevState) {
-    const replaceChoicesHaveChanged = prevProps.replaceChoices !== this.props.replaceChoices;
+    const hasReplaceChoicesChanged = prevProps.replaceChoices !== this.props.replaceChoices;
     if (
-      prevState.codeMirrorMode !== this.state.codeMirrorMode ||
-      replaceChoicesHaveChanged
+      prevState.codeMirrorMode !== this.state.codeMirrorMode || hasReplaceChoicesChanged
     ) {
       // Changed from code mirror mode to read only mode or vice versa,
       // so setup the other editor.
--- a/lib/components/helpers/pretty-text-input.js
+++ b/lib/components/helpers/pretty-text-input.js
@@ -41,9 +41,8 @@ export default React.createClass({
 
   componentDidUpdate: function(prevProps, prevState) {
     const hasReplaceChoicesChanged = prevProps.replaceChoices !== this.props.replaceChoices;
-    if (
-      prevState.codeMirrorMode !== this.state.codeMirrorMode || hasReplaceChoicesChanged
-    ) {
+    const hasCodeMirrorModeChanged = prevState.codeMirrorMode !== this.state.codeMirrorMode;
+    if (hasCodeMirrorModeChanged || hasReplaceChoicesChanged) {
       // Changed from code mirror mode to read only mode or vice versa,
       // so setup the other editor.
       this.createEditor();
--- a/lib/components/helpers/pretty-text-input.js
+++ b/lib/components/helpers/pretty-text-input.js
@@ -40,7 +40,9 @@ export default React.createClass({
   },
 
   componentDidUpdate: function(prevProps, prevState) {
-    if (prevState.codeMirrorMode !== this.state.codeMirrorMode) {
+    const wasPreviewing = prevProps.field.meta.isPreviewing;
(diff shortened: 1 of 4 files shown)


lib/components/helpers/pretty-text-input.js
```diff
@@ -40,7 +40,16 @@ export default React.createClass({
   },
 
   componentDidUpdate: function(prevProps, prevState) {
-    if (prevState.codeMirrorMode !== this.state.codeMirrorMode) {
+    const wasPreviewing = prevProps.field.meta.isPreviewing;
```
jdeal: The `meta` property is only available when the meta plugin is enabled. So you can't depend on that being here. And if you look around, you'll see that none of the formatic's components know about the meta plugin at all.

Beyond that, I think "preview" is a zapier-ism, and sniffing a change of that value to "kick" codemirror is kind of smelly.

I think the real problem is that you're not seeing codemirror update when you change the value, and we should probably just fix that. If there are perf problems with that, we'll just have to involve a debounce or something to tame it.

lib/components/helpers/pretty-text-input.js
```diff
@@ -40,7 +40,11 @@ export default React.createClass({
   },
 
   componentDidUpdate: function(prevProps, prevState) {
-    if (prevState.codeMirrorMode !== this.state.codeMirrorMode) {
+    const replaceChoicesHaveChanged = prevProps.replaceChoices !== this.props.replaceChoices;
```
jdeal: Really minor thing, but I making the boolean word a prefix, so maybe `hasChangedReplaceChoices` instead?

lib/components/helpers/pretty-text-input.js
```diff
@@ -40,7 +40,11 @@ export default React.createClass({
...
+    if (
+      prevState.codeMirrorMode !== this.state.codeMirrorMode ||
+      replaceChoicesHaveChanged
```
jdeal: I think @stevemolitor knows better than me what's going on here, but I wonder if we could just call `tagCodeMirror` instead of totally re-creating CodeMirror? If we do have to re-create it, we probably want to test that with a `setInterval` changing `replaceChoices` and see how that acts. I'm thinking it will probably lose focus if that happens. Which may not generally be a problem, but it could cause some weirdness when the choices first load.

stevemolitor: Do only the the replace choice labels change, not the values? If the values can change, there might be an issue with just re-tagging CodeMirror. For example, let's say the initial choices are:

```js
[ 
  { value: 'fname', label: 'First Name'},
  { value: 'lname', label: 'Last Name'},
]
```

And the text box value is `{{fname}} {{lname}}`. So it renders "First Name Last Name" with pretty pills.

If the `fname` choice is then changed (or its value changed):

```js
// new choices
[ 
  { value: 'lname', label: 'Last Name'},
]
```

Then it _should render `{{fname} Last Name`, since `fname` is no longer a valid tag. I'm guessing that's not gonna happen though?

Another issue though is what if we're debouncing code mirror when this happens. To be safe we should probably do something similar to `onChangeAndTagCodeMirror`, except we don'' always want to fire the `onChange`. I think.

So, assuming only the labels change, yeah I think we can re-tag code mirror, making sure we don't lose debounced changes, or just re-render the read-only editor if we're not in CodeMirror mode. 

Golly isn't this a fun component?

efhjones: @stevemolitor not sure if I follow:
>Then it should render {{fname} New Last Name, since fname is no longer a valid tag.

I think it would just stick with the previous tag, rather than switching to the next available tag? AFAICT it'll interpret the tag based on [what value is there](https://github.com/zapier/formatic/blob/master/lib/components/helpers/pretty-text-input.js#L401), and since the interpretations for that tag would have disappeared, it'll just stick with the old tag value.

@jdeal just calling `tagCodeMirror` doesn't work -- looks like it fails here on [`getValue`](https://github.com/zapier/formatic/blob/master/lib/components/helpers/pretty-text-input.js#L367) though I'm having a hard time investigating further with `npm link` not working like it used to.

Calling `onChangeAndTagCodeMirror` causes an infinite loop, not sure if that's what you meant by avoiding calling `onChange` @stevemolitor . I tried a new function hoping that would avoid the `getValue` issue I was having earlier, but this only works if the user hasn't updated/interacted with the field:

```
    if (!this.codeMirror) {
      // We might have erased our codemirror instance before hitting the trailing
      // end of the debounce. If so, get the value out of state.
      if (this.props.value !== this.state.value) {
        this.onChange(this.state.value);
      }
      return;
    }
    this.tagCodeMirror();
```
@jdeal what's your main concern with creating a new editor every time? I don't expect `replaceChoices` to be changing very often, but I could be wrong.

efhjones: @jdeal I tested out with `setInterval` as you mentioned, it does lose focus when it makes new choices, but I'm curious what the root of the concern is there? Do `replaceChoices` change all that often?

jdeal: @efhjones It's not that we change choices that often. I was more thinking of a case that might be kind of common where someone clicks into a field before the choices have loaded, and then once they load, they get kicked out of the field. But I _think_ there's currently a bigger problem there which is that the choices might not load at all. So this PR should at least improve things in that respect. So I'll go ahead an approve for now.

I do think problems with `npm link` should be separated from whether or not we merge a PR. So grab somebody to help you out with that.

efhjones: @jdeal I can see the reason for concern re: 
>  they get kicked out of the field
It could definitely happen if there were a _ton_ of `replaceChoices` , so that they're super slow to load and the user is already editing a field at the point. I'm not sure of a good way around that, but I'll investigate.

Could you elaborate a bit on the problem re:
> the choices might not load at all

Should formatic be concerned about that? As far as these current changes it's only listening for a change, I don't think it's dependent on `replaceChoices` being filled out or accurate. Would you mind giving a little context on the problems that would create? Thanks!

jdeal: @efhjones I meant that if you click into a field before `replaceChoices` have loaded, then CodeMirror won't update after the choices have loaded, which means it might not be properly tagged. Again, I'm not an expert on the flow here, so @stevemolitor might have a better understanding of what would happen there. Or you can just try it out: use a `setTimeout` to load choices slowly, click in the field, and then see if the tags work properly.

efhjones: I double check this today @jdeal , it looked to me like everything was tagged correctly. I do get kicked out of the field, but I set a timeout to load the choices after 5 seconds, so I typed a few chars in the field next to the tag and it looked like everything went ok. 

@stevemolitor took a look a the perf implications of this change here and things look pretty promising. We investigated some other strategies to update the code without creating a new editor, but we hit a few dead ends there, and we doubt that it would improve performance from what we've experienced here. However, I did find some changes that I can make in `zapier` that won't cause formatic to rerun unless necessary, so I think perf has actually improved on that standpoint.

stevemolitor:
@efhjones and I discussed this in slack, but to summarize:

Running `tagCodeMirror()` again doesn't help, because CodeMirror won't update unless the _text_ changes. It's trying to optimize / reduce 'unnecessary' renders. So as long as the value passed to CodeMirror is ie `{{fname}} {{lname}}`, it will never re-render. 

@efhjones tried the `changed()` (sp?) method in CodeMirror but couldn't get it working. 

We discussed and rejected some very kludgy workarounds involving forced changes to the text. ;)

efhjones:
Thanks for the summary @stevemolitor ! I'll merge this on Monday unless there are any other reservations.

lib/default-config.js
```diff
@@ -1100,21 +1100,20 @@ export default function (config) {
       // Array of choice arrays should be flattened.
       choices = _.compact(_.flatten(choices));
 
-      choices.forEach(function (choice, i) {
+      const choicesWithLabels = choices.map(function (choice) {
         // Convert any string choices to objects with `value` and `label`
```
efhjones: Previously, iterating over these and mutating the original `choices` array would throw an error if passed an immutable object. This resolves that by not mutating the items in the `choices` array, but providing a new array with modifications, if necessary. If all the choices are not a string and have a label, this should return a copy of the original array.

stevemolitor: Bah, mutation bad, functional good. 👍 !

PR #84, opened 2017-10-01
Add node-libs-browser dev dependency

MikeTheReader:
With a fresh install the `npm start` command throws the following error:

```
Error: Cannot find module 'node-libs-browser'
      at Function.Module._resolveFilename (module.js:470:15)
      at Function.Module._load (module.js:418:25)
      at Module.require (module.js:498:17)
      at require (internal/module.js:20:19)
 at Object.<anonymous> (/home/michael/Projects/javascript/react/formatic/node_modules/webpack/lib/node/NodeSourcePlugin.js:7:23)
```

Running `npm install --save-dev node-libs-browser` takes care of this. This PR just commits the addition of the dependency to package.json. I followed the pattern of setting hard version (no ^).

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "jquery": "2.1.4",
     "lodash": "3.10.1",
     "marked": "0.3.6",
+    "node-libs-browser": "^2.0.0",
     "open": "0.0.5",
     "portscanner": "2.1.1",
     "react": "15.3.0",
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "jquery": "2.1.4",
     "lodash": "3.10.1",
     "marked": "0.3.6",
-    "node-libs-browser": "^2.0.0",
+    "node-libs-browser": "2.0.0",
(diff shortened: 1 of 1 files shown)


jdeal:
Hmm, @MikeTheReader, I wonder why I've never seen that. 🤔

MikeTheReader:
@jdeal Not sure. I was able to replicate it on two machines (both using Linux). I did find that Webpack 1.10.2 includes `node-libs-browser` as a dependency (https://github.com/webpack/webpack/commit/232618e489ae477d744cbd8a00327fb3c00852c4), so modifying that in the package.json also resolves this. If I'm the only one seeing this, though, I can certainly get rid of this pull request.

jdeal:
Ah, I wonder if it's just that I have an older version of `npm` that's installing the peer dependency vs a newer version of `npm` for you. Since it's a peer dependency, that makes sense eithere way. I'll go ahead and merge. Thanks!

PR #85, opened 2017-10-12
Migrate to React 16

ngryman:
Hey guys,

So this migrates Formatic to React 16 :tada:

A bunch of dependencies had to be upgraded/replaced as a side-effect but hopefully, everything's working.

To sum up:

- `react-addons-test-utils` → `react-dom/test-utils`
- `React.createClass` is no more → `create-react-class`
- `React.PropTypes` is no more → `prop-types`.
- `react-addons-css-transition-group` → `react-transition-group` (1.x).
- `react-addons-update` → `immutability-helper`.
- (optional) add a `requestAnimationFrame` polyfill to supress warnings.
- (optional) `react-hot-loader` → `react-hot-loader/webpack`.

@jdeal FYI the bug was related to the fact that we have to use `react-transition-group` 1.x for a drop-in replacement (oops).

Tests are passing and the demo seems ok, but please make a quick double check as my knowledge of this project is pretty limited for now. Thanks!

### DIFF
--- a/lib/components/helpers/pretty-text-input.js
+++ b/lib/components/helpers/pretty-text-input.js
@@ -4,6 +4,7 @@
 /*eslint no-script-url:0 */
 
 import React from 'react';
+import createReactClass from 'create-react-class';
 import ReactDOM from 'react-dom';
 import cx from 'classnames';
 
@@ -29,7 +30,7 @@ var toString = function (value) {
    instantiated when the user moves the mouse into the edit area.
    Initially a read-only view using a simple div is shown.
  */
-export default React.createClass({
+export default createReactClass({
 
   displayName: 'PrettyTextInput',
 
--- a/lib/components/helpers/pretty-text-input.js
+++ b/lib/components/helpers/pretty-text-input.js
@@ -10,7 +10,7 @@ import cx from 'classnames';
 
 import TagTranslator from './tag-translator';
 import _ from '../../undash';
-import {keyCodes} from '../../utils';
+import {keyCodes, ref} from '../../utils';
 import HelperMixin from '../../mixins/helper';
 
 var toString = function (value) {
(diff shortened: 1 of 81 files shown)


jdeal:
@ngryman This seems pretty solid overall. I compared all the demo fields from the current version to this upgrade and everything is working except one weird edge with the pretty text field, where the cursor is acting a little weird.

If you click in the whitespace on the first line in the old version, your cursor goes to the left of the pill:



but in the new version, the cursor ends up in the middle of the pill:



If you use the arrow keys to move left and right, you'll also see that in the new version, you can't get to the end of the field like you can in the old version.

No idea what's going on there. We do some funky stuff there with codemirror, and I know the new version of React has some now portal stuff. So maybe some weird interactions there. Maybe we could use the new portal stuff there and that would solve it? Not sure.

jdeal:
Oh, also, can you clean up those limiting errors before merging?

lib/components/helpers/pretty-text-input.js
```diff
@@ -376,7 +376,12 @@ export default createReactClass({
       });
     };
 
-    this.codeMirror.operation(tagOps);
+    // XXX: Make sure we apply those operation after React has made its rendering pass.
+    // As React 16 is asynchronous and uses rAF, it's same to assume that this will
+    // be called after React has patched the DOM.
+    requestAnimationFrame(() => {
```
ngryman: @jdeal It's kind of hacky solution but it works well.
As we are rendering tags outside the React rendering loop, we want to make sure our tags have been committed to the DOM so CodeMirror can safely do its internal stuff.

ngryman: A cleaner solution would be to get rid of the current system and render everything in the `renderDefault` function switching between the read-only editor and CodeMirror. But I didn't want to mess around with this as it already works.

jdeal: Ooh, nice fix! I was thinking that was going to be nastier. And yeah, I really, really, really would like to switch to Draft.js so we could just render like normal.

PR #89, opened 2017-12-01
Bump all devDependencies

vitorbal:
Bumps all devDependencies to latest, which also makes sure uglify-js gets updated.


vitorbal:
Hmm I'm not sure I know what you mean, @ngryman. I bumped all dev dependencies to latest, and not just latest minor. Is there any dependency in specific I might have missed?

The formatting issues were caught by eslint after updating, which is why I had to fix them here (otherwise the PR would be failing)

vitorbal:
I tested all npm scripts and everything seems to work fine, so this should be safe to merge. Not sure it warrants a release by itself though.

ngryman:
I don't exactly know what those upgrades bring to the table. But I don't see any reason why not to release in order to bring good stuff to consumers.

Given the relatively slow pace of development of this project and the high coupling with `zapier/zapier`, I think it makes sense to release as often as possible.

vitorbal:
Agreed with all your points, @ngryman! However, this PR doesn't have any user-facing changes, since it only updates dev packages, so in the end, I decided to not cut a new release for it.

PR #91, opened 2017-12-06
compare values in replaceChoices rather than object

mariyadiminsky:
Because it's possible to pass in an identical array with a different ref, we should check if the contents of `replaceChoices` is the same rather than doing a straight `!==` comparison. We want to avoid unnecessarily calling `this.createEditor` unless we have new `replaceChoices`. Without using `_.isEqual`, we'll trigger on all changes to `replaceChoices`, even if the contents of the array itself don't change. This can cause us to lose focus in a field we're typing if `replaceChoices` is rebuilt on render.

### DIFF
--- a/lib/components/helpers/pretty-text-input.js
+++ b/lib/components/helpers/pretty-text-input.js
@@ -41,7 +41,7 @@ export default createReactClass({
   },
 
   componentDidUpdate: function(prevProps, prevState) {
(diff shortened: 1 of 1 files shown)


efhjones:
@stevemolitor would you mind giving this a quick review? Changes are pretty small, this should fix a fairly nasty bug.

PR #92, opened 2017-12-11
Remember cursor when insert clicked

efhjones:
https://github.com/zapier/formatic/pull/83 introduced a bug that would lose the current location of the cursor if the `replaceChoices` changed when the insert button was clicked, causing nodes inserted into the CodeMirror editor to appear in unexpected places. This PR remembers the location of the cursor just before the insert button is clicked so we retain that position when `handleChoiceSelection` is called.


lib/components/helpers/pretty-text-input.js
```diff
@@ -357,7 +363,9 @@ export default createReactClass({
 
     var textBox = this.textBoxRef;
     textBox.innerHTML = ''; // release any previous read-only content so it can be GC'ed
-
+    this.setState({
```
jdeal: @efhjones Did you try setting this synchronously and have problems? In other words, just setting a local variable and then setting the position after creating codemirror? It looks like there's some possible async business in `tagCodeMirror`, but you could just pass the position along to that function and set the position after `this.codeMirror.operation(tagOps)`.

lib/components/helpers/pretty-text-input.js
```diff
@@ -111,7 +112,13 @@ export default createReactClass({
       if (pos) {
         this.codeMirror.replaceRange(tag, {line: pos.line, ch: pos.start}, {line: pos.line, ch: pos.stop});
       } else {
+        if (this.state.cursorPosition) {
+          this.codeMirror.setCursor(this.state.cursorPosition);
```
jdeal: It seems odd to me to only set the position when picking a choice. Seems like we should set the position right after creating the new codemirror instance. See my other comment.

lib/components/helpers/pretty-text-input.js
efhjones: No I didn't have any problems @jdeal but was worried about that so I was hoping to get/set it only when needed. However your suggestion above makes sense! I'll rework it, thanks!

PR #95, opened 2018-01-12
Improve accessibility of Pretty Text Input field

efhjones:
We have event listeners for `onTouchStart` and `onMouseEnter`, but that means we're not listening for `onFocus`, which happens when a user attempts to tab into the field. Without this listener the field won't switch to codemirror when someone tries to tab into the field, so let's add that listener to improve our accessibilty.

@jdeal Would you mind taking a look at this? Not sure if `onFocus` was left off for other reasons


lib/components/helpers/pretty-text-input.js
```diff
@@ -284,7 +284,10 @@ export default createReactClass({
     // Render read-only version.
     return (
       <div onKeyDown={this.onKeyDown} className={cx({'pretty-text-wrapper': true, 'choices-open': this.state.isChoicesOpen})}
-        onMouseEnter={this.switchToCodeMirror} onTouchStart={this.switchToCodeMirror}>
+        onMouseEnter={this.switchToCodeMirror}
+        onTouchStart={this.switchToCodeMirror}
+        onFocus={this.switchToCodeMirror}
```
WestonThayer: Have you tested if VoiceOver navigation triggers focus?

efhjones: @WestonThayer It will if you tab onto the field and then hit `ctrl + optn + shift + downarrow`, but the voiceover just calls it a 'group', which isn't very informative 😅
Adding `onFocus` made it so that when I tabbed onto the field it said 'text area' instead.

WestonThayer: `onFocus` sounds like a win! Thanks so much for figuring this out so quickly 😀 

I'd love to help QA this, would it be possible to deploy to a greek server or staging?

efhjones: @WestonThayer No problem! Hopefully it's the right fix, I'm not sure if we left `onFocus` off for a reason.

I'll ping you in slack as far as QAing this! 😄

efhjones:
Awesome, thanks @WestonThayer !

PR #96, opened 2018-01-17
Remove Old Files

xavdid:
Spring cleaning for some old files. Makes the whole thing look more professional.


xavdid:
We could also remove `bower.json` since that service is mostly getting shutdown. There's instructions [here](https://bower.io/blog/2017/how-to-drop-bower-support/).

PR #98, opened 2018-03-01
Make array safe for non-array values.

stevemolitor:
Shouldn't happen, but if the Array component is passed a non-array
value, turn it into an array of one element, so that the `forEach` in
getInitialState doesn't blow up.

### DIFF
--- a/lib/components/fields/array.js
+++ b/lib/components/fields/array.js
@@ -28,7 +28,11 @@ export default createReactClass({
 
     var items = this.props.field.value;
     if (!Array.isArray(items)) {
-      items = [items];
+      if (items !== null && items !== undefined) {
+        items = [items];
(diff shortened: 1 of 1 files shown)


lib/components/fields/array.js
```diff
@@ -27,6 +27,9 @@ export default createReactClass({
     var lookups = [];
 
     var items = this.props.field.value;
+    if (!Array.isArray(items)) {
+      items = [items];
```
jdeal: I wonder if we should check if `items == null` and in that case make an empty array? I just think `[null]` and `[undefined]` wouldn't make much sense.

stevemolitor:
Changed to check for null / undefined as per @jdeal's comment.

PR #99, opened 2018-06-07
Include requiredText as a child

michaelgilley:
Allows for a new, optional config key on field that will render the actual `required-text` inside the required label `<span />`. Doing this allows for custom required text to be rendered and a new className is added for additional css targeting in such cases.

### DIFF
--- a/lib/components/helpers/label.js
+++ b/lib/components/helpers/label.js
@@ -26,6 +26,12 @@ export default createReactClass({
     var config = this.props.config;
     var field = this.props.field;
     var fieldLabel = config.fieldLabel(field);
+    var fieldIsRequired = config.fieldIsRequired(field);
+    var requiredTextClassName = cx({
+      'required-text': fieldIsRequired,
+      'not-required-text': !fieldIsRequired,
+      'has-required-text': !!field.required_text,
+    });
     var label = null;
 
     if (typeof this.props.index === 'number') {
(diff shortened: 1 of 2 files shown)


jdeal:
👀

lib/components/helpers/label.js
```diff
@@ -26,6 +26,12 @@ export default createReactClass({
     var config = this.props.config;
     var field = this.props.field;
     var fieldLabel = config.fieldLabel(field);
+    var fieldIsRequired = config.fieldIsRequired(field);
+    var requiredTextClassName = cx({
+      'required-text': fieldIsRequired,
+      'not-required-text': !fieldIsRequired,
+      'has-required-text': !!field.required_text,
```
jdeal: Note how we don't access the `field.required` property directly. That's a pattern to avoid binding the components directly to the fields and instead allow the `config` to override that. So, instead, we should add a method to the config so you can do `config.fieldRequiredText(field)` here.

For the `default-config` implementation, return `field.requiredText` just because Formatic uses camelCased properties. (It allows snake_case for helpText but nevermind that.)

But if we provide the same exact text for every field, the plugin on the Zapier could just be implemented as:

```js
config.fieldRequiredText = () => 'This field is required.';
```

jdeal:
Actually, hang on please, I want to chat about this first.

michaelgilley:
Closing this in favor of #100

PR #100, opened 2018-06-07
Add and use RequiredLabel factory in Label

michaelgilley:
Replace the hardcoded `<span />` for displaying required text in a field with it's own factory element so it can be overridden/extended in a plugin.

### DIFF
--- a/lib/components/helpers/required-label.js
+++ b/lib/components/helpers/required-label.js
@@ -1,0 +1,38 @@
+// # required label component
+
+/*
+  Required Label for a field
+*/
+
+'use strict';
+
+import React from 'react';
+import createReactClass from 'create-react-class';
+import cx from 'classnames';
+
+import HelperMixin from '../../mixins/helper';
+
+export default createReactClass({
+
+  displayName: 'RequiredLabel',
+
+  mixins: [HelperMixin],
+
+  render: function () {
+    return this.renderWithConfig();
+  },
+
+  renderDefault: function () {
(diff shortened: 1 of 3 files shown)


lib/components/helpers/required-label.js
```diff
@@ -0,0 +1,38 @@
+// # required label component
+
+/*
+  Required Label for a field
+*/
+
+'use strict';
```
jdeal: Weird, not sure why we have all those here. `¯\_(ツ)_/¯`

lib/components/helpers/required-label.js
```diff
@@ -0,0 +1,38 @@
...
+  displayName: 'RequiredLabel',
+
+  mixins: [HelperMixin],
+
+  render: function () {
+    return this.renderWithConfig();
+  },
+
+  renderDefault: function () {
+    var config = this.props.config;
```
jdeal: I know Formatic is a mix of styles, but you can feel free to use `const` and `let` if you want. We need to do a big cleanup pass, so either way is fine.

lib/components/helpers/required-label.js
michaelgilley: 🤷‍♂️

lib/components/helpers/required-label.js
michaelgilley: Yeah wasn't sure. When in Rome? I'll just leave this for now and it can be refactored later with all the rest.

PR #101, opened 2018-07-06
Add CustomChoice Display Value config option

michaelgilley:
Adds a config option to override the Display Value of a field when Custom Choice is selected. Before the field would just displaying the choice's label but this option defaults to the label, allowing plugins to override this behavior and customize the displayed value.

### DIFF
--- a/lib/components/helpers/pretty-select-value.js
+++ b/lib/components/helpers/pretty-select-value.js
@@ -258,11 +258,9 @@ export default createReactClass({
 
     if (this.state.isEnteringCustomValue || (!isDefaultValue && !currentChoice && currentValue)) {
       if (this.hasCustomField()) {
-        const {choices} = this.props;
+        const {choices, config} = this.props;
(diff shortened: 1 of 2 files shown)


lib/components/helpers/pretty-select-value.js
```diff
@@ -258,11 +258,9 @@ export default createReactClass({
 
     if (this.state.isEnteringCustomValue || (!isDefaultValue && !currentChoice && currentValue)) {
       if (this.hasCustomField()) {
-        const {choices} = this.props;
+        const {choices, config} = this.props;
         const customChoice = _.find(choices, (choice) => choice.action === 'enter-custom-value');
-        if (customChoice && customChoice.label) {
-          return customChoice.label;
-        }
+        return config.customChoiceDisplayValue(customChoice, currentValue) || currentValue;
```
arkwright: The only thing I can think of to improve this is potentially refactoring the `|| currentValue`, as it duplicates the `return currentValue` below.

PR #102, opened 2018-07-12
plugin hint boxes

jdeal:
This adds a button to each example in the demo that shows "hint" boxes around each component for each `createElement_` plugin method. This gives some visual indication of where each method is used and which methods are available for each field type.

plugin-hints


vitorbal:
oooh this is pretty neat! Thanks for this @jdeal!

PR #108, opened 2018-09-11
Add support for field messages. This will enable us to show inline er…

stevelikesmusic:
…ror messaging via plugins

This adds a hook to fields to inject a message below fields' children (which would be the input field). Using this new plugin hook and overriding individual createElement hooks for error classes, we can get colorized inline messaging.

### DIFF
--- a/lib/components/helpers/field.js
+++ b/lib/components/helpers/field.js
@@ -96,6 +96,7 @@ export default createReactClass({
       >
         {label}
         {help}
+        {config.renderFieldMessage(this.props)}
       </div>
     );
   },
--- a/lib/default-config.js
+++ b/lib/default-config.js
@@ -405,6 +405,11 @@ export default function(config) {
       return config.renderComponent(component);
     },
 
+    // Render a message below the field.
+    renderFieldMessage: function() {
+      return null;
+    },
+
     // Normalize an element name.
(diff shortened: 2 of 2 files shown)


stevelikesmusic:
I'm not sure overriding `help` would work in this case. `help` in `field.js` is the combined `helpText` and input `children`. So, we'd still need to insert the message after `help` and its `children`. 

```js
const help = config.cssTransitionWrapper(
  this.state.collapsed
     ? []
     : [
	   // overriding this still leaves out children. the message needs to appear
	   // below children.
       config.createElement('help', {
         config,
         field,
         key: 'help',
       }),
       this.props.children,
     ]
);
```

but is this more of what you're thinking for option 2?

```js
const fieldMessage = config.createElement('fieldMessage', {
  config,
  field,
  key: 'fieldMessage'
})

return (
  <div
    className={cx(classes)}
    style={{ display: field.hidden ? 'none' : '' }}
  >
    {label}
    {help}
    {fieldMessage}
  </div>
);
```

jdeal:
Ah, I see what you're saying @stevelikesmusic. The code is a bit misleading there.

I think there's another way to do this with the existing code though. Just override `createElement_Field`. You can intercept the child (second prop) and wrap it.

PR #113, opened 2019-02-20
Docs overhaul + light refactoring + build improvements + misc cleanup

jdeal:
This is the "make Formatic slightly less crufty" refactor. Hopefully this will set the stage for a later "make Formatic hardly crufty at all" refactor. Or maybe even a "make Formatic modern with hooks and other goodie" refactor. Anyway...

The main goal here was to update the docs which were quite out of date and out of sync with the local demo. A secondary goal was to overhaul the way docs were built from the old very custom solution to a new less-custom Next.js solution. Along the way, lots of other updating, both large and small, took place.

- [x] Replace the custom docs build with a Next.js build.
- [x] Update babel and some other dependencies.
- [x] Move the webpack-based demo over to Next.js and merge it so the demo and the docs are all one thing.
- [x] Add enzyme for testing components.
- [x] Add a test utility to easily mount some fields for testing.
- [x] Add a `FieldContainer` to make it easier to plugin components for handling field types.
- [x] Add a teensy bit of documentation for `FieldContainer`.
- [x] Add snapshot testing for all fields in the demo.
- [x] Add a new `renderTag` plugin method so field/helper components can pass rendering of all tags through the plugin method. This way, literally any part of a field/helper component can be overridden. The main reason to do this is to be able to override styling for the demo in a not-global-CSS-soup way.
- [x] Change all the field/helper components to render everything through the `renderTag` plugin method.
- [x] Add a CSS plugin that spits out some BEM-like (but not really BEM) class names.
- [x] Change the formatic.css file to use those new classes for styling the demo, making the demo slightly less ugly.
- [x] Rewrite/update all the docs and examples.
- [x] Rewrite all the doc components, including removing the reliance on Bootstrap.
- [x] Add emotion to use for writing the docs pages/components. (Only a dev dependency.)
- [x] Change the rollup config to spit out dual builds, one for CommonJS and one for ES modules.
- [x] Add `"sideEffects": false` to `package.json` so webpack can start tree-shaking things. Only `cssPlugin` and `FieldContainer` can currently be tree-shaken, but see next point...
- [x] Export all plugins and mixins so that we can deprecate the non-tree-shakable things hanging off the default exported Formatic component.
- [x] Make it so CodeMirror can be used from a dependency vs having to hang off of `window`.
- [x] Remove any remnants of bower. Yuck!
- [x] Remove Bootstrap. Bye!
- [x] Remove gulp and just rely on npm scripts.
- [x] Remove other unused junk.
- [x] Breaking change: Swap the order of children rendered in array items to match object items and to better support flexbox styling. This is the main thing that requires a major version bump. But lots of other stuff is changing, so it's better to just call this a major bump anyway. A `renderTag` plugin can be used to swap the order back to the original order.
- [x] Fix conflicts.
- [x] Test with Zapier to make sure nothing blows up.


FormaticTestUtils.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { mount } from 'enzyme';
```
stevelikesmusic: Any thought to using `react-testing-library`?

LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2014 Zapier
+Copyright (c) 2018 Zapier Inc.
```
stevelikesmusic: Should this be 2019 now :trollface:

__tests__/components/field-container.js
```diff
@@ -0,0 +1,42 @@
...
+                onChangeValue(event.target.value.toUpperCase())}
+              value={props.field.value}
+            />
+          )}
+        </FieldContainer>
+      ),
+    }));
+    const fields = [{ type: 'upper', key: 'name' }];
+    const value = { name: 'foo' };
+    const onChangeFormSpy = jasmine.createSpy('onChangeForm');
```
stevelikesmusic: Why call `jasmine` instead of `jest`?

__tests__/components/fields/array.js
```diff
@@ -0,0 +1,10 @@
+/*global describe, it, expect*/
+
+import { renderFieldsToHtml } from '../../../FormaticTestUtils';
```
stevelikesmusic: If we're sticking with `rollup` any want or thoughts of using the (resolve alias plugin](https://github.com/rollup/rollup-plugin-alias)

docs/server/getSnippets.js
```diff
@@ -0,0 +1,30 @@
+/* global __dirname */
+import fs from 'mz/fs';
```
stevelikesmusic: I'm full of questions 😄. Do we still need `mz` now that `util.promisify` gets us essentially the same thing?

docs/server/snippets/plugin-fieldLabel.js
```diff
@@ -1,11 +1,13 @@
-var plugin = function (config) {
-
+const plugin = () => {
   return {
-    fieldLabel: function (field) {
+    fieldLabel: function(field) {
```
stevelikesmusic: Any reason to not use an arrow function here?

docs/server/snippets/plugin-using-multiple.js
```diff
@@ -0,0 +1,12 @@
+//CUT
```
stevelikesmusic: This realization was one of my Formatic "ah ha" moments when it finally clicked what the lib could do.

FormaticTestUtils.js
jdeal: I hadn't but maybe I should. It just wasn't another rabbit hole I wanted to go down right now.

LICENSE
jdeal: Yep, that's how long it took to finish this PR. 😅

__tests__/components/field-container.js
jdeal: No good reason. Switched it now.

__tests__/components/fields/array.js
jdeal: Totally cool by me. I think I'll hold off for another PR for it. Maybe you're volunteering? 😄 If I understand, it means the tests use the rollup babel plugin? I hadn't thought about doing that and wasn't aware of that alias plugin either. In other projects, I've just used a symlink for tests: https://github.com/jdeal/qim/tree/master/test/node_modules.

docs/server/getSnippets.js
jdeal: Good point. Done.

docs/server/snippets/plugin-fieldLabel.js
jdeal: Nope, probably just copied over from somewhere else. Fixed. I'd like to go through the codebase and switch lots of things to arrow functions as well, but had to cut things off somewhere. 😄

docs/server/snippets/plugin-using-multiple.js
jdeal: Yeah, it makes Formatic plugins really powerful. But it can also confuse people too. 😬

__tests__/components/fields/array.js
stevelikesmusic: I'm either/or—no strong feeling as to how. But it might be nice to skip relative paths 🤷‍♂️

docs/server/snippets/plugin-fieldLabel.js
stevelikesmusic: I get that!

lib/formatic.js
```diff
@@ -26,10 +26,31 @@ import ScrollMixin from './mixins/scroll.js';
...
-import referencePlugin from './plugins/reference';
+import cssPlugin from './plugins/css-plugin';
 import elementClassesPlugin from './plugins/element-classes';
 import helpTextPlacementPlugin from './plugins/help-text-placement';
+import metaPlugin from './plugins/meta';
+import referencePlugin from './plugins/reference';
+
+export { default as FieldContainer } from './components/field-container';
+
+export {
```
jdeal: Effectively, we're deprecating the old way of grabbing these off of `Formatic`.

lib/formatic.js
```diff
@@ -26,10 +26,31 @@ import ScrollMixin from './mixins/scroll.js';
...
+  bootstrapPlugin,
+  cssPlugin,
```
jdeal: This new plugin is only available as a named export.

lib/mixins/shared.js
```diff
@@ -0,0 +1,54 @@
...
+
+  onBlurAction: function() {
+    this.onStartAction('blur');
+  },
+
+  isReadOnly: function() {
+    return this.props.config.fieldIsReadOnly(this.props.field);
+  },
+
+  renderWith: function(elementName, extraProps) {
```
jdeal: Create an object that points to `config.renderTag`, so the `jsx` function can render through the config.

bower.json
```diff
@@ -1,33 +0,0 @@
-{
```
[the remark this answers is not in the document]
stevelikesmusic: 😆

docs/server/getSnippets.js
jdeal: I actually killed off all this special snippet API stuff anyway and just use `raw-loader` to load in snippets. TLDR: for a static site, Next will still try to run getInitialProps on the client, but there's no server, so of course it fails.

PR #117, opened 2019-02-23
Use alias to get rid of ../ paths for modules

jdeal:
Lots of ../../.. in the source and in the docs made things pretty cumbersome. Now, there's an alias `@` which points to the root directory. I went with `@` instead of a name like `root` just to avoid any confusion thinking there's a real npm module with that name.

Also, this could have been done with a single babel plugin, but there seems to be some concern that jest or webpack caching may break, so instead I went with tool-specific plugins/settings for webpack/jest/rollup. See https://github.com/tleunen/babel-plugin-module-resolver/issues/346 and other issues for details on the potential problem with the babel plugin.

For now, I kept all the `./` sibling paths. I'd be find switching those to absolute, but there isn't as much to gain there, so I just left them for now.

cc @stevelikesmusic I'll be merging this right away to keep things moving, but feel free to continue the convo if you have any other opinions here.


stevelikesmusic:
Nice addition @jdeal!

PR #120, opened 2019-02-27
Make sure we cancel the trailing debounced invocation of CM changes

ngryman:
Codemirror is regularly crashing on production: https://sentry.io/organizations/zapier-1/issues/370070265/events/4de4e354523046cdbe8a295e7aefcfdc/.

Given the `Cannot read property 'getValue' of null` error, it's very likely due to the fact that we debounce an event handler that is invoked after we got rid of the CM instance.

This PR leverage `Function#cancel` provided by `_.debounce` in order to cancel any trailing debounced invocation. This allow us to safely assume that every event handler will have a valid CM instance and help to simplify things a little bit.

I took the liberty to simplify a couple of location (cf. inline comments), but I would definitely love some experienced 👀 on this to make sure I'm not doing anything stupid.

### DIFF
--- a/src/components/helpers/pretty-text-input.js
+++ b/src/components/helpers/pretty-text-input.js
@@ -83,7 +83,7 @@ export default createReactClass({
   componentWillReceiveProps: function(nextProps) {
     // If we're debouncing a change, then we should just ignore this props change,
     // because there will be another when we hit the trailing edge of the debounce.
-    if (this.isDebouncingCodeMirrorChange) {
+    if (this.debouncedOnChangeAndTagCodeMirror) {
       return;
     }
 
@@ -438,7 +438,7 @@ export default createReactClass({
     //
     // But if we are calling this function from CodeMirror itself, we want to stay
     // in sync with it's internal state.
-    if (this.isDebouncingCodeMirrorChange) {
+    if (this.debouncedOnChangeAndTagCodeMirror) {
       this.maybeCodeMirrorOperation(tagOps);
       this.maybeSetCursorPosition(cursorPosition);
     } else {
--- a/src/components/helpers/pretty-text-input.js
+++ b/src/components/helpers/pretty-text-input.js
@@ -520,8 +520,12 @@ export default createReactClass({
     this.codeMirror.off('focus', this.onFocusCodeMirror);
 
     if (this.debouncedOnChangeAndTagCodeMirror) {
+      // Cancel any trailing invocation
       this.debouncedOnChangeAndTagCodeMirror.cancel();
       this.debouncedOnChangeAndTagCodeMirror = null;
+
+      // Flush changes
+      this.onChange(this.codeMirror.getValue());
     }
 
     this.codeMirror = null;
--- a/src/components/helpers/pretty-text-input.js
+++ b/src/components/helpers/pretty-text-input.js
@@ -179,14 +179,7 @@ export default createReactClass({
   },
 
   onBlur() {
-    if (this.isDebouncingCodeMirrorChange) {
-      this.onChangeAndTagCodeMirror();
-      this.isDebouncingCodeMirrorChange = false;
-      this.setState({ hasFocus: false }, this.props.onBlur);
-    } else {
-      this.setState({ hasFocus: false });
-      this.props.onBlur();
-    }
+    this.setState({ hasFocus: false }, this.props.onBlur);
   },
 
   insertBtn: function() {
@@ -409,11 +402,16 @@ export default createReactClass({
 
     const textBox = this.textBoxRef;
     textBox.innerHTML = ''; // release any previous read-only content so it can be GC'ed
-    const pos = this.codeMirror ? this.codeMirror.getCursor() : null;
+
     this.codeMirror = this.props.config.codeMirror()(textBox, options);
     this.codeMirror.on('change', this.onCodeMirrorChange);
     this.codeMirror.on('focus', this.onFocusCodeMirror);
 
+    this.debouncedOnChangeAndTagCodeMirror = _.debounce(() => {
+      this.onChangeAndTagCodeMirror();
+    }, 200);
+
+    const pos = this.codeMirror ? this.codeMirror.getCursor() : null;
     this.tagCodeMirror(pos);
   },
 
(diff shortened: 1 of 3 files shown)


ngryman:
👇 Annotations 👇

src/components/helpers/pretty-text-input.js
```diff
@@ -179,14 +179,7 @@ export default createReactClass({
...
   onBlur() {
-    if (this.isDebouncingCodeMirrorChange) {
-      this.onChangeAndTagCodeMirror();
-      this.isDebouncingCodeMirrorChange = false;
-      this.setState({ hasFocus: false }, this.props.onBlur);
-    } else {
-      this.setState({ hasFocus: false });
-      this.props.onBlur();
-    }
+    this.setState({ hasFocus: false }, this.props.onBlur);
```
ngryman: I assumed that in both case we can invoke `this.props.onBlur` using the `Component#setState` callback, it helps to simplify the code a little bit.

Another assumption I made is that we don't necessarily need to invoke `onChangeAndTagCodeMirror` here as it will be invoked shortly after the `blur` event and I don't see any outstanding issue of doing so, but I'm not 💯 sure of that.

src/components/helpers/pretty-text-input.js
```diff
@@ -409,11 +402,16 @@ export default createReactClass({
...
     textBox.innerHTML = ''; // release any previous read-only content so it can be GC'ed
-    const pos = this.codeMirror ? this.codeMirror.getCursor() : null;
+
     this.codeMirror = this.props.config.codeMirror()(textBox, options);
     this.codeMirror.on('change', this.onCodeMirrorChange);
     this.codeMirror.on('focus', this.onFocusCodeMirror);
 
+    this.debouncedOnChangeAndTagCodeMirror = _.debounce(() => {
+      this.onChangeAndTagCodeMirror();
+    }, 200);
```
ngryman: Create the debounced version of `onChangeAndTagCodeMirror` when we create the editor instead of doing it on the fly in `onCodeMirrorChange`. This is mainly to better reason about setup/teardown code related to CM.

src/components/helpers/pretty-text-input.js
```diff
@@ -452,14 +450,6 @@ export default createReactClass({
...
 
   onChangeAndTagCodeMirror() {
-    if (!this.codeMirror) {
-      // We might have erased our codemirror instance before hitting the trailing
-      // end of the debounce. If so, get the value out of state.
-      if (this.props.value !== this.state.value) {
-        this.onChange(this.state.value);
-      }
-      return;
-    }
```
ngryman: This won't be invoked anymore when the CM instance has died. I'm not sure of the exact implication of not invoking `onChange` here anymore though. I've done a QA locally and to me, everything seemed ok. But again, not sure I didn't miss something.

src/components/helpers/pretty-text-input.js
```diff
@@ -531,10 +512,18 @@ export default createReactClass({
...
     textBoxNode.removeChild(cmNode);
+
     this.codeMirror.off('change', this.onCodeMirrorChange);
     this.codeMirror.off('focus', this.onFocusCodeMirror);
+
+    if (this.debouncedOnChangeAndTagCodeMirror) {
+      this.debouncedOnChangeAndTagCodeMirror.cancel();
+      this.debouncedOnChangeAndTagCodeMirror = null;
+    }
+
```
ngryman: Make sure we cancel any trailing invocation so we are safe everywhere else.

stevemolitor:
@ngryman did you qa locally on zapier/zapier in the editor? If so can you remind me of the `yarn link` incantations again? ;) Thanks!

src/components/helpers/pretty-text-input.js
jdeal: Uh, hmm, how do I tell you this... the `_.debounce` here is not actually lodash's debounce. It's Formatic's "undash" debounce: https://github.com/zapier/formatic/blob/48ceba41b2c04917e1ac1449dfadf77544d725bc/src/undash.js#L74-L108

It's a fake lodash with just the bare necessities. So if you want to be able to call `cancel`, you'll have to change the implementation. Which is pretty straightforward.

src/components/helpers/pretty-text-input.js
jdeal: Yeah, always calling `onBlur` in `setState` seems fine. Hopefully there are no dragons there.

I _think_ the other change is problematic though, and it's related to the bug you're fixing here. If you blur away and changes are being debounced, if the field gets unmounted before the debounce, we'll lose the changes. This commit seems to be intentionally fixing something, and I think that's what it was, since I remember that problem happening: https://github.com/zapier/formatic/commit/11ebdb58e1bd42d2b58c33f94fa2e06d0fbc9d00.

src/components/helpers/pretty-text-input.js
jdeal: Pretty sure you can't just remove this. I'm not quite sure when this would happen and not blur per the other problem I mentioned, but it's the same kind of thing I was fixing in this commit:

https://github.com/zapier/formatic/commit/00d0766324afd47f0fe5e52e67cafdf4b08b6f38

If the component gets unmounted while debouncing, we lose the value that's in the state, which means that value won't get saved by the surrounding app.

src/components/helpers/pretty-text-input.js
ngryman: Haha I got baited 😄 Thanks for pointing it out, I will implement the `cancel`.

src/components/helpers/pretty-text-input.js
ngryman: Oh gotcha! So if I understand correctly we could hit a edge case where the component is umounted within the 200ms timespan after the user blurs the field. If so we effectively lose the last batch of input...

src/components/helpers/pretty-text-input.js
```diff
@@ -531,10 +512,22 @@ export default createReactClass({
...
     this.codeMirror.off('change', this.onCodeMirrorChange);
     this.codeMirror.off('focus', this.onFocusCodeMirror);
+
+    if (this.debouncedOnChangeAndTagCodeMirror) {
+      // Cancel any trailing invocation
+      this.debouncedOnChangeAndTagCodeMirror.cancel();
+      this.debouncedOnChangeAndTagCodeMirror = null;
+
+      // Flush changes
+      this.onChange(this.codeMirror.getValue());
```
ngryman: @jdeal I think that we should flush changes when we are about to destroy the CM editor and cancel debounced events so we don't have to manage edge cases in `onBlur` or `onChangeAndTagCodeMirror`.

jdeal: Yep, I think it makes a lot more sense to do this here @ngryman!

PR #123, opened 2019-03-18
♿️Add major ARIA support to most fields

ruebel:
## Overview

This PR aims to start the ball rolling for better ARIA support in formatic.

## Features

* Adds better label / input support for most of our input types
* Fixes most ARIA issues in docs caused by supporting docs elements

## Notes

🚧 This PR does not take care of all ARIA issues but it significantly reduces the amount of issues.

## Screenshots from Docs Field Types page

Before:

After:

### DIFF
--- a/__tests__/components/fields/object.js
+++ b/__tests__/components/fields/object.js
@@ -122,6 +122,7 @@ describe('object field', () => {
 
   it('should render changes from outside', () => {
     let value = {
+      id: 'object-test',
       object: {
         key1: 'value1',
         key2: 'value2',
@@ -135,6 +136,8 @@ describe('object field', () => {
       formatic,
       ObjectClass
     );
+
+    delete object.state.id;
 
     expect(object.state).toEqual({
       assocList: [
@@ -151,6 +154,8 @@ describe('object field', () => {
     };
 
     renderToNode(value);
+
+    delete object.state.id;
 
     expect(object.state).toEqual({
       assocList: [
@@ -175,6 +180,8 @@ describe('object field', () => {
       formatic,
       ObjectClass
     );
+
+    delete object.state.id;
 
     expect(object.state).toEqual({
       assocList: [
@@ -207,6 +214,8 @@ describe('object field', () => {
     };
 
     renderToNode(value);
+
+    delete object.state.id;
 
     // no change:
     expect(object.state).toEqual({
--- a/__tests__/components/fields/object.js
+++ b/__tests__/components/fields/object.js
@@ -137,14 +137,10 @@ describe('object field', () => {
       ObjectClass
     );
 
-    delete object.state.id;
-
-    expect(object.state).toEqual({
-      assocList: [
-        { key: 'key1', value: 'value1' },
-        { key: 'key2', value: 'value2' },
-      ],
-    });
+    expect(object.state.assocList).toEqual([
+      { key: 'key1', value: 'value1' },
+      { key: 'key2', value: 'value2' },
+    ]);
 
     value = {
       object: {
@@ -155,14 +151,10 @@ describe('object field', () => {
 
     renderToNode(value);
 
-    delete object.state.id;
-
-    expect(object.state).toEqual({
-      assocList: [
-        { key: 'key2', value: 'value2' },
-        { key: 'keyNew', value: 'value1' },
-      ],
-    });
+    expect(object.state.assocList).toEqual([
+      { key: 'key2', value: 'value2' },
+      { key: 'keyNew', value: 'value1' },
+    ]);
   });
 
   it('should ignore changes from outside when there are duplicate keys', () => {
@@ -181,14 +173,10 @@ describe('object field', () => {
       ObjectClass
     );
 
-    delete object.state.id;
-
-    expect(object.state).toEqual({
-      assocList: [
-        { key: 'key1', value: 'value1' },
-        { key: 'key2', value: 'value2' },
-      ],
-    });
+    expect(object.state.assocList).toEqual([
+      { key: 'key1', value: 'value1' },
+      { key: 'key2', value: 'value2' },
+    ]);
 
     renderToNode(value);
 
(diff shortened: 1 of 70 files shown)


ruebel:
@jdeal I'm sure this is low priority on your TODO list so no rush but I just wanted to make sure that this review request wasn't lost in your pile of things from when you were on vacation

jdeal:
@ruebel This is on my todo list! Unfortunately, yeah, it was buried under a stack of things, but I promise I will get to it as soon as I can. If it's a blocker for something, let me know, and I'll try to reprioritize.

__tests__/components/fields/object.js
```diff
@@ -136,6 +137,8 @@ describe('object field', () => {
       ObjectClass
     );
 
+    delete object.state.id;
```
jdeal: What if you just change the test to use `object.state.assocList` since that's all it uses from the state anyway?

docs/components/Page.js
```diff
@@ -39,6 +40,12 @@ const styles = {
 
 const Page = props => (
   <Layout pageKey={props.pageKey} pages={pages}>
+    <Head>
+      <title>
+        Formatic
+        {props.title ? ` | ${props.title}` : ''}
```
jdeal: What about using `pages[props.pageKey].title` to avoid the duplicated strings?

docs/components/Section.js
```diff
@@ -22,7 +22,11 @@ const Section = props => (
   <div css={styles.section}>
     <h2 css={styles.sectionHead} id={dashify(props.title)}>
       {props.title}
-      <a css={styles.link} href={`#${dashify(props.title)}`}>
+      <a
+        aria-label="anchor link"
```
jdeal: Maybe this could be "example link" to give it a bit more context?

pages/demo.js
```diff
@@ -270,7 +270,7 @@ class FormDemo extends Component {
 const sortedExamples = _.sortBy(examples, ['title']);
 
 const DemoPage = () => (
-  <Page pageKey="demo">
+  <Page pageKey="demo" title="Field Types">
```
jdeal: See comment above: I think you can just pull this from the `pages` object in the `Page` component.

pages/plugins.js
```diff
@@ -10,7 +10,7 @@ import Code from '@/docs/components/Code';
 import { loadSnippets } from '@/docs/utils';
 
 const Plugins = props => (
-  <Page pageKey="plugins">
+  <Page pageKey="plugins" title="Plugins">
```
jdeal: Ditto above.

pages/start.js
```diff
@@ -30,7 +30,7 @@ const basicExampleValue = {
 };
 
 const Start = props => (
-  <Page pageKey="start">
+  <Page pageKey="start" title="Getting Started">
```
jdeal: Ditto above.

src/components/fields/checkbox-array.js
```diff
@@ -83,12 +86,13 @@ export default createReactClass({
             <span key="spacer" renderWith={this.renderWith('InputSpacer')}>
               {' '}
             </span>
-            <span
+            <label
```
jdeal: Just make sure to test this on the Zapier side just in case anything is depending on this markup. Hopefully there's no evil `span` selector, but better to be safe. I guess it's murky whether this is a major release given it _technically_ has a breaking change here. But you could also call it a patch since you're fixing the accessibility issues. I'll leave that decision to you.

src/components/helpers/assoc-list-item-key.js
```diff
@@ -36,6 +36,7 @@ export default createReactClass({
 
     return (
       <input
+        aria-labelledby={this.props.id ? `${this.props.id}_label` : undefined}
```
jdeal: Hmm, is this accurate? Is it appropriate to have multiple inputs be labeled by a single label? It seems like something like `aria-label="key"` would be more appropriate here? And `aria-labelledby` should apply to the widget as a whole?

src/components/helpers/assoc-list-item-value.js
```diff
@@ -32,7 +32,11 @@ export default createReactClass({
     const field = this.props.field;
 
     const fieldElem = config.createFieldElement({
-      field,
+      field: {
+        ...field,
+        id: `${this.props.id}_${this.props.index}`,
+      },
+      labelledby: this.props.id ? `${this.props.id}_label` : undefined,
```
jdeal: Same question as the key above. Is it appropriate to have multiple fields labeled by the same label?

src/mixins/field.js
```diff
@@ -11,6 +11,18 @@ import shared from './shared';
 export default {
   ...shared,
 
+  // Create a unique id for the field
+  getInitialState: function() {
+    return {
+      id:
+        this.props.field.id ||
+        Math.random()
```
jdeal: I'll state what is probably obvious: that for a _huge_ list of fields, this could mysteriously replicate. It's probably not a significant enough risk to worry about though.

ruebel: I feel like if you have a form with that many fields you should be tried for crimes against humanity, but I am definitely open to suggestions on this if there is a better way to generate an id

src/components/helpers/assoc-list-item-key.js
ruebel: Yeah I wasn't sure what to do here, I think your idea is better 👍

src/components/fields/checkbox-array.js
ruebel: I've never tested formatic in z/z, do we have a guide?  If not would you like to pair on that some time to make sure I'm not exploding anything?

jdeal: Primarily it's just a matter of `yarn link`, but yeah, happy to pair on more specifics!

ruebel:
@jdeal I did a pass of local testing with Zapier and things were looking good but would still like to sync up with you sometime to make sure I'm not missing something.

PR #125, opened 2019-04-19
Add isDisabled to single-line-string

ruebel:
✨ Adds `isDisabled` flag to the `single-line-string` field type.
📸 Updates snapshots to match

### DIFF
--- a/demo/examples/single-line-string.js
+++ b/demo/examples/single-line-string.js
@@ -28,6 +28,14 @@ const fields = [
   },
   {
     id: 'single-line-string-4',
+    label: 'Disabled Name',
+    type: 'single-line-string',
+    key: 'readonlyName',
+    default: 'Robert',
+    isDisabled: true,
+  },
+  {
+    id: 'single-line-string-5',
     label: 'Path',
     type: 'single-line-string',
     key: 'path',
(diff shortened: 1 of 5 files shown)


src/components/fields/single-line-string.js
```diff
@@ -49,6 +49,7 @@ export default createReactClass({
         autoComplete={field.autoComplete}
         autoFocus={field.autoFocus}
         className={cx(this.props.classes)}
+        disabled={field.isDisabled}
```
jdeal: I prefer the `is` prefix but when in Rome, I’d say to follow the convention of not using the prefix for field properties.

demo/examples/single-line-string.js
```diff
@@ -28,6 +28,14 @@ const fields = [
   },
   {
     id: 'single-line-string-4',
+    label: 'Disabled Name',
+    type: 'single-line-string',
+    key: 'readonlyName',
```
jdeal: I know it’s just a demo but do we maybe want to name this `disabledName` to avoid confusion in the demo?

src/components/fields/single-line-string.js
jdeal: To match readOnly, we probably should also wrap in a plug-in method, but we can always wrap later if needed too.

ruebel: I think I did the plug-in method correctly, but double-check me on this one

PR #126, opened 2019-04-27
Add AutoFields. Move field-components to their own folder.

stevelikesmusic:
## Description

Adds an AutoFields component. The idea is to use a hook to get the value of the form. In this way, either explicitly pass in `AutoFields` as a child of `FormContainer`, or leave `FormContainer` empty to get the same result:

```jsx
<FormContainer onChange={onChange} value={formValues}>
  <AutoFields />
</FormContainer>
```

is the same as 

```jsx
<FormContainer onChange={onChange} value={formValues} />
```
One thing to note is how we're inferring the schema. In this first pass it's not ideal in that we infer the schema with in the component function. As a result, we'll infer the types of values each time `onChange` is fired. One alternative would be to `useMemo` to get the schema from values, but this could be problematic if we have a dynamic form that adds new values (which should add form fields). The most performant way might be to use a schema if we think we'll be using dynamic forms. Then if we're inferring the schema from values, we can assume the number of fields will always be static. 

We'll also need to pass a schema to do some more interesting things with `AutoFields` such as mixing it with static fields.


__tests__/future/AutoFields.js
```diff
@@ -0,0 +1,72 @@
...
+describe('AutoFields', () => {
+  test('renders fields from defaultValues', () => {
+    const onChangeSpy = jest.fn();
+    const { getByLabelText } = render(
+      <ExampleForm defaultValue={defaultValue} onChange={onChangeSpy} />
+    );
+    fireEvent.change(getByLabelText('First Name'), {
+      target: { value: 'Joe' },
+    });
+    expect(onChangeSpy).toHaveBeenNthCalledWith(1, {
```
stevelikesmusic: Found this method to be easier to read than `spy.mock.calls[0][0]`

__tests__/future/AvoidRerendering.js
```diff
@@ -38,4 +38,35 @@ describe('avoid rerendering', () => {
       ['firstName'],
     ]);
   });
+
+  test('avoids unnecessary renders of AutoFields', () => {
```
stevelikesmusic: Just to ensure `AutoFields` matches the same rendering behavior as other form fields.

src/future/ReactiveValue.js
```diff
@@ -23,6 +23,10 @@ const createReactiveValue = defaultValue => {
   }
 
   function getCurrentValueAt(key) {
+    if (key === undefined) {
```
stevelikesmusic: So we can get the value for the whole form in `AutoFields` without passing down props.

src/future/formatic-fields/index.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
```
stevelikesmusic: Moved this all out of `/index.js` to group all of our form field components.

src/future/formatic-fields/index.js
```diff
@@ -0,0 +1,31 @@
...
+    );
+  }
+  FieldComponent.displayName = `${fieldType}Field`;
+  return FieldComponent;
+}
+
+export const TextField = createField('Text', TextInput);
+export const IntegerField = createField('Text', IntegerInput);
+
+export const AutoFields = createAutoFields({
```
stevelikesmusic: Using a factory here to avoid any circular dependencies since `AutoFields` requires other field components.

src/future/index.js
```diff
@@ -35,7 +33,7 @@ export function FormContainer({
         onChange={onChange}
         value={isControlled ? value : savedDefaultValue}
       >
-        {children}
+        {children || <AutoFields />}
```
stevelikesmusic: If no children, let's use `AutoFields` similarly to classic Formatic.

src/future/formatic-fields/index.js
jdeal: Works for me. Curious: why `formatic-fields` and not just `fields`?

src/future/ReactiveValue.js
jdeal: I also just realized, I think there's a more ominous problem here. I think there's a mismatch between the value you're getting and what's being subscribed to. At the very least, this makes the code hard to follow. But worse, there will likely be memory leaks or improper notifications. I think we should keep value based subscriptions separate from key based subscriptions so it's easy to follow what's happening and avoid these problems. We'll probably also want to add some tests at some point to guarantee subscriptions are cleaned up properly. Assuming we can do that without sneaking into implementation details. 🤔

stevelikesmusic: Ahh good points @jdeal. I was thinking just about getting the values of the form, but I agree, this also opens a can of worms.

src/future/index.js
```diff
@@ -35,7 +34,7 @@ export function FormContainer({
         onChange={onChange}
         value={isControlled ? value : savedDefaultValue}
       >
-        {children}
+        {children || <AutoFields components={components} />}
```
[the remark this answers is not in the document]
stevelikesmusic: I wonder if there'd be value in having a `Formatic` export that includes this behavior? Would that make transitions to the new version a little easier? Maybe not if using a code mod—it'd be just as easy to add an `AutoFields` component as the child.

src/string-utils.js
```diff
@@ -0,0 +1,14 @@
+export function capitalize(string = '') {
```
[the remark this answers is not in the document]
stevelikesmusic: I always think of uppercased camelCase names as `ComponentCase`. Any time I see that naming convention in a React project that doesn't use `*.jsx`, I think "This is a component".  Would you be fine with `stringUtils.js`? What about even having a folder `utils` and then dropping `utils` from the file name? `/utils/string.js`?

demo/future/AutoFields.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+
+import { FormContainer, useField } from '@/src/future';
+
+export function CustomTextField({ fieldKey, label }) {
```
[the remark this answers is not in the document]
stevelikesmusic: Thanks—I hadn't dug it out of the code just yet. Yeah I think we'll want to live test in a browser

src/future/formatic-fields/index.js
stevelikesmusic: My thought was we'll probably have a `Fields` component here, and it might be confusing to have `/fields/Fields.js`.

jdeal: Ah, gotcha. That works then. Maybe something like `builtin-fields` would work too? Just thinking out loud, `formatic-fields` is fine, just wondering if others might wonder what it means too.

stevelikesmusic: I'm good with `builtin-fields`. Just didn't want it to be simply `fields`.

src/future/index.js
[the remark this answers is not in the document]
jdeal: Maybe? I didn't specify this, but I'm thinking right now, we push a "clean" version as far as we can go, and then we think about what that means to existing Formatic usage after that. Do we use a codemod or an adapter or just a "good luck"? I'm not sure yet, but I think if we worry too much about the transition now, it might dirty up the rewrite. Or at least make it harder on our brains. 😄

src/string-utils.js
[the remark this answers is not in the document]
jdeal: I'm fine with whatever here. Just to explain though... I use `PascalCase` for namespace/package names. I actually got that convention from `React` which is just a namespace/package, not a class or a component. And in turn, I try to match file names to how they would be imported. So:

```js
import * as StringUtils from './StringUtils`
```

Even though typically you wouldn't import everything like that, that's how I think of it. I'm fine letting you set the precedent here though, and I'll follow. This is definitely minutia.

src/future/index.js
```diff
@@ -24,17 +21,15 @@ export function FormContainer({
...
   const [savedDefaultValue] = useState(defaultValue);
   const [isControlled] = useState(savedDefaultValue === undefined);
+  const normalizedValue = isControlled ? value : savedDefaultValue;
   // Keep context object pure.
-  const renderContext = useMemo(() => ({ renderTag, renderComponent }), [
-    renderTag,
-    renderComponent,
-  ]);
+  const renderContext = useMemo(
+    () => ({ renderTag, renderComponent, initialValue: normalizedValue }),
```
jdeal: Passing the `initialValue` through the `RenderContext` like this seems a bit strange to me. I think it would be better to just have this available from `ReactiveValue` as `useInitialReactiveValue`. I'm going to make some changes to `ReactiveValue` though, so this is fine as-is, and I can clean it up then.

jdeal:
@stevelikesmusic I went ahead and merged. In case I conflict in my changes, I didn't want to cause conflicts in your branch.

PR #127, opened 2019-05-09
Notify meta on value

stevelikesmusic:
Fixes a minor bug with `setValue`. We now update meta when the root value is updated.

Additional commits are noted. But one add is `jest-dom` which includes some convenience [matchers](https://github.com/testing-library/jest-dom#table-of-contents) to `expect`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,1 +1,18 @@
-module.exports = require('zapier-scripts/src/config/eslint');
+module.exports = {
+  extends: './node_modules/zapier-scripts/src/config/eslint.js',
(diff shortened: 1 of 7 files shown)


.eslintrc.js
```diff
@@ -1 +1,18 @@
-module.exports = require('zapier-scripts/src/config/eslint');
+module.exports = {
```
stevelikesmusic: This will get yanked once we get rid of all the `createReactClass` calls

stevelikesmusic:
Good call @jdeal! Thanks for that 👍

PR #128, opened 2019-05-10
Do not fire the onChange event in pretty-text-input when no changes.

stevemolitor:
Do not fire the onChange event in pretty-text-input helper when there are no changes waiting to be debounced. Doing so can cause bugs in client applications that depend on onChange only being fired when the user actually changes something.

### DIFF
--- a/src/components/helpers/pretty-text-input.js
+++ b/src/components/helpers/pretty-text-input.js
@@ -535,7 +535,9 @@ export default createReactClass({
       this.debouncedOnChangeAndTagCodeMirror = null;
 
       // Flush changes
-      this.onChange(this.codeMirror.getValue());
+      if (this.isDebouncingCodeMirrorChange) {
+        this.onChange(this.codeMirror.getValue());
+      }
     }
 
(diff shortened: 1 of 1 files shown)


stevemolitor:
This may have been introduced here: https://github.com/zapier/formatic/commit/9f6a9898dac0eab348fc47ef1a9e38511b070c94.

I will add some tests before asking for a review.

ngryman:
Oh I remember this one 😱 I think I may be missing some context now. Do you mind adding a test that illustrate the issue?

stevemolitor:
> I think you're description of the change sounds bigger than the change itself. You're just making sure we don't flush changes if there are no changes to flush when unmounting, right? If you can confirm, I'll go ahead and approve this.

Yep that's it! That unnecessary flush can cause `onChange` to fire, which can cause formatic clients to mark something as 'dirty' when it's not, causing .... problems!

stevemolitor:
I kind of dropped the ball on this, so to get this moving and fixing the underlying bug in the client app, I'm going to merge.

PR #129, opened 2019-05-29
update code test snapshot

stevemolitor:
Fix failing snapshot test.


stevemolitor:
@jdeal Have you seen this error? Locally when running `npm run patch-release` the snapshot test failed because the snapshot expected ie `autocorrect="false"` but received `autocorrect="off"`. So I updated the snapshot to expect `"off"`. But in travis it receives `"false". So I can make test test pass locally but not in travis, or in travis but not locally, but not both. 😱 

I did nuke `node_modules` and did `npm install` locally just to rule that out.

jdeal:
@stevemolitor I've never seen this, and your branch fails for me locally. No idea why you're seeing different results. Just to get you moving, ping me Monday, and I can do the patch-release.

PR #130, opened 2019-07-04
Limit height of choices accordion to fit within its container

efhjones:
### Description

This PR shortens the max height at which an accordion can open. 

Previously, it could expand to be the entire height of its container (from the top of the containing rect).

This meant that the dropdown would often expand past the viewbox and users would have to scroll down to see all the options in the dropdown.

This PR fixes that by making the max height 75% of the height of its container, so that it is always contained within the viewport.

### DIFF
--- a/src/components/helpers/choices.js
+++ b/src/components/helpers/choices.js
@@ -256,7 +256,7 @@ export default createReactClass({
       const clampedHeight = Math.min(Math.abs(propsMaxHeight), 1);
       const actualHeight = windowHeight - top;
       const desiredMaxHeight = actualHeight * clampedHeight;
-      if (actualHeight !== desiredMaxHeight) {
+      if (Math.round(actualHeight) !== Math.round(desiredMaxHeight)) {
         didSetState = true;
         this.setState(
           {
--- a/src/components/helpers/choices.js
+++ b/src/components/helpers/choices.js
@@ -253,7 +253,7 @@ export default createReactClass({
       const clampedRatio = Math.min(Math.abs(maxHeightRatio), 1);
       const actualHeight = windowHeight - top;
       const desiredMaxHeight = actualHeight * clampedRatio;
-      if (Math.round(actualHeight) !== Math.round(desiredMaxHeight)) {
+      if (Math.round(desiredMaxHeight) !== Math.round(this.state.maxHeight)) {
         didSetState = true;
         this.setState(
           {
--- a/src/components/helpers/choices.js
+++ b/src/components/helpers/choices.js
@@ -249,7 +249,7 @@ export default createReactClass({
       const rect = node.getBoundingClientRect();
       const top = rect.top;
       const windowHeight = window.innerHeight;
-      const height = windowHeight - top;
+      const height = (windowHeight - top) * 0.75;
       if (height !== this.state.maxHeight) {
         didSetState = true;
         this.setState(
--- a/src/components/helpers/choices.js
+++ b/src/components/helpers/choices.js
@@ -249,12 +249,18 @@ export default createReactClass({
       const rect = node.getBoundingClientRect();
       const top = rect.top;
       const windowHeight = window.innerHeight;
-      const height = (windowHeight - top) * 0.75;
-      if (height !== this.state.maxHeight) {
+      const propsMaxHeight =
+        this.props.field && this.props.field.meta
+          ? this.props.field.meta.choicesMaxHeight
+          : 1;
+      const clampedHeight = Math.min(Math.abs(propsMaxHeight), 1);
+      const actualHeight = windowHeight - top;
+      const desiredMaxHeight = actualHeight * clampedHeight;
+      if (actualHeight !== desiredMaxHeight) {
         didSetState = true;
         this.setState(
           {
-            maxHeight: height,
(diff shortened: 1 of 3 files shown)


hoonpark:
https://zapierorg.atlassian.net/browse/CORE-1770

efhjones:
@stevemolitor Good questions, and thanks for the review, I'm glad to have your opinion on this! 

> Does this affect formatic dropdowns outside the editor? If so is everyone ok with those changes?

It does, so I'm hesitant to make this change. We could very well have people out there who prefer the current behavior 🤔 

> If the above is a concern, can we plugin in the magic 0.75 value, or do it with a max-height css value or something?

We can -- though the manner feels a bit hacky to me-- we'll could pass it through the `field.meta` object. It feels strange to me to accept changes from `this.props.field.meta` but since this is such a nested component I'm not sure how else to pass the prop for our use case.

I've got another working solution pushed now. I'm not completely pleased with it but it does work 🤔 It's a bit complicated because this component already uses JS to manage the `maxHeight` by evaluating the size of the viewport: 

https://github.com/zapier/formatic/blob/bcc6bb0064191c8a07378663c4fd6e5c38a86e75/src/components/helpers/choices.js#L253-L262

So here's the potential solution I've come up with in 

Via a prop passed in `meta`:

1. See if the prop `choicesMaxHeight` exists in the `field.meta`. It must be a decimal between 0 and 1.
1. If so, ensure it's above 0 and less than 1, default to 1
1. Multiply the actual height by the desired height (calculated in the previous step)
1. If the actual height and desired height are not equal, set the maxHeight of the element

Let me know what you think of this, thanks again @stevemolitor !

src/components/helpers/choices.js
```diff
@@ -249,12 +249,18 @@ export default createReactClass({
       const rect = node.getBoundingClientRect();
       const top = rect.top;
       const windowHeight = window.innerHeight;
-      const height = windowHeight - top;
-      if (height !== this.state.maxHeight) {
+      const propsMaxHeight =
```
stevemolitor: @efhjones Doesn't sound too bad. I'm guessing that we will want the same max height for all choices dropdowns in the editor? If so it might be easier to pass `choicesMaxHeight` via the formatic config:

```js
  const configMaxHeight = this.props.config.getChoicesMaxHeight;
  //...
```

In the [default formatic config](https://github.com/zapier/formatic/blob/master/src/default-config.js) you could provide the default implementation of `getPropsMaxHeight` that would return `1`:

```js
  getChoicesMaxHeight {
    return 1;
  }
```

Then in our `formaticFieldsConfig` we could specify `0.75`. That way the editor will use that value, but nothing else changes.

Minor thing: I'm not sure if `choicesMaxHeight` is the best name here. That makes me think it's the actual max height, like in pixels or something. But actually it's the ratio of available height to actual max height. So... (drawing a blank here...) .. uhm `choicesMaxHeightRatio` or something? Not totally clear, but at least it doesn't lull me into thinking it's the actual max height.

src/components/helpers/choices.js
```diff
@@ -249,12 +249,18 @@ export default createReactClass({
...
+        this.props.field && this.props.field.meta
+          ? this.props.field.meta.choicesMaxHeight
+          : 1;
+      const clampedHeight = Math.min(Math.abs(propsMaxHeight), 1);
+      const actualHeight = windowHeight - top;
+      const desiredMaxHeight = actualHeight * clampedHeight;
+      if (actualHeight !== desiredMaxHeight) {
```
stevemolitor: I wonder if we'll get weird rounding errors, where `actualHeight !== desiredMaxHeight` when really they are the same, because of JS weirdness with floating points? If so we can round both values before comparing.

efhjones: 👍 Great catch!

src/components/helpers/choices.js
efhjones: > If so it might be easier to pass choicesMaxHeight via the formatic config

That's an excellent point! Thanks! 

> That makes me think it's the actual max height, like in pixels or something

Great point as well! I'm struggling to come up with something better than `choicesMaxHeightRatio` so we'll go with that! 😄

efhjones:
Alrighty @stevemolitor the latest changes are ready for another look whenever you have time. Thanks so much for the great suggestions!

src/components/helpers/choices.js
```diff
@@ -249,12 +249,15 @@ export default createReactClass({
       const rect = node.getBoundingClientRect();
       const top = rect.top;
       const windowHeight = window.innerHeight;
-      const height = windowHeight - top;
-      if (height !== this.state.maxHeight) {
+      const maxHeightRatio = this.props.config.getChoicesMaxHeightRatio();
+      const clampedRatio = Math.min(Math.abs(maxHeightRatio), 1);
```
stevemolitor: This will blow up if `config.getChoicesMaxHeightRatio` is not defined, and it won't be by default. I think we should add a default implementation to the [default formatic config](https://github.com/zapier/formatic/blob/master/src/default-config.js)

```js
  getChoicesMaxHeight {
    return 1;
  }
```

The default formatic config also kind of serves as documentation, letting you know what can be customized.

efhjones:
Thanks @stevemolitor !

PR #131, opened 2019-10-17
Fix bug trying ta tag non-existing CodeMirror instance.

stevemolitor:
Sometimes we see this error in the console:

This happens when the debounced method to tag the CodeMirror editor tries tag this.codeMirror after the component has unmounted and this.codeMirror has been set to null on componentWillUnmount. To fix, bail out of the method if `this.codeMirror` is not defined.

Fixing this might reduce some sentry errors.

### DIFF
--- a/src/components/helpers/pretty-text-input.js
+++ b/src/components/helpers/pretty-text-input.js
@@ -417,7 +417,12 @@ export default createReactClass({
 
     this.debouncedOnChangeAndTagCodeMirror = _.debounce(() => {
       this.isDebouncingCodeMirrorChange = false;
-      this.onChangeAndTagCodeMirror();
+
+      // If this.codeMirror no longer exists don't tag, to avoid this error:
+      // https://cdn.zapier.com/storage/photos/69d375365045c5fd6e9c7e2a68c43905.png
+      if (this.codeMirror) {
+        this.onChangeAndTagCodeMirror();
+      }
     }, 200);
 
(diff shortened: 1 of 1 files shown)


src/components/helpers/pretty-text-input.js
```diff
@@ -417,7 +417,12 @@ export default createReactClass({
 
     this.debouncedOnChangeAndTagCodeMirror = _.debounce(() => {
       this.isDebouncingCodeMirrorChange = false;
-      this.onChangeAndTagCodeMirror();
+
+      // If this.codeMirror no longer exists don't tag, to avoid this error:
+      // https://cdn.zapier.com/storage/photos/69d375365045c5fd6e9c7e2a68c43905.png
+      if (this.codeMirror) {
+        this.onChangeAndTagCodeMirror();
```
jdaudier: By "tag", do we mean create a pretty pill?

stevemolitor: Yes. We create a [`PrettyTag`](https://github.com/zapier/formatic/blob/fix-bug-tagging-non-existent-code-mirror/src/components/helpers/pretty-tag.js#L19) element [here](https://github.com/zapier/formatic/blob/fix-bug-tagging-non-existent-code-mirror/src/components/helpers/pretty-text-input.js#L440-L444), and then tell CodeMirror to use that whenever it sees a certain `{{123__value}}` text value.

PR #136, opened 2020-03-30
Allow ID to be passed to pretty-select-field

efhjones:
This PR allows an ID to be passed through to pretty-select-fields.


efhjones:
@stevemolitor Any chance you might have a few minutes to take a gander here? 🙏 This just allows us to utilize testing IDs if we want to for the e2e tests -- make the elements easier to find 😄 Thank you!

ngryman (approved):
I don't know why but I got a notification from GH about this PR ¯\_(ツ)_/¯
The change is pretty straightforward so I think I can ✅ this.

efhjones:
👍 Thanks so much @ngryman !

